PR #60, opened 2024-04-29
Waypoint: update the list display to use a table

catsby:
Simple change to how we list templates from just dumping them out to tabling them.

Ex:

templates_table

### DIFF
--- a/internal/commands/waypoint/templates/list.go
+++ b/internal/commands/waypoint/templates/list.go
@@ -76,5 +76,11 @@ func listTemplates(opts *TemplateOpts) error {
 		templates = append(templates, resp.GetPayload().ApplicationTemplates...)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/commands/waypoint/templates/list.go
```diff
@@ -76,5 +76,11 @@ func listTemplates(opts *TemplateOpts) error {
 		templates = append(templates, resp.GetPayload().ApplicationTemplates...)
 	}
 
-	return opts.Output.Show(templates, format.Pretty)
+	templateFields := []format.Field{
```
paladin-devops, line 79: I'm good with reducing the # of fields displayed to just these 3! Makes sense that we give the user the name of the template to look up with `read` if they want more info.

paladin-devops (approved):

PR #63, opened 2024-05-07
[WAYP-2571] TFC Config error

HenryEstberg:

This is a small bugfix to return a better error message when a user attempts to read a Waypoint TFC Config and one has not been set.

### How I've tested this PR:

This has been manually tested and the acceptance tests have been run.

Run `make go/test` to run the acceptance tests

To manually test, build the binary and connect to your HCP org. If that org already has a TFC Config set for waypoint, first run:

```
hcp waypoint tfc-config delete
```

To test the change, run:

```
hcp waypoint tfc-config read
```

### Output of affected commands:
--> 
```
% hcp waypoint tfc-config read  
X No TFC Config found for this project

```

### DIFF
--- a/internal/commands/waypoint/tfcconfig/tfc_config_read.go
+++ b/internal/commands/waypoint/tfcconfig/tfc_config_read.go
@@ -5,6 +5,7 @@ package tfcconfig
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/client/waypoint_service"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/models"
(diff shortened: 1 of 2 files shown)


internal/commands/waypoint/tfcconfig/tfc_config_read.go
```diff
@@ -67,6 +67,9 @@ func readRun(opts *TFCConfigReadOpts) error {
 	if err != nil {
 		return errors.Wrapf(err, "%s failed to get TFC Config", opts.IO.ColorScheme().FailureIcon())
 	}
+	if resp.Payload.TfcConfig == nil {
+		return errors.New("No TFC Config found for this project")
```
briancain, line 71: Is this an error though? I think given that we updated our API to say no TFC config isn't an error, this simply returns 0.

briancain (changes requested):

internal/commands/waypoint/tfcconfig/tfc_config_read.go
catsby: What Brian said; I think we should just output "no tfc config found" or something.

HenryEstberg: Sure we can just output that message without calling it an error.

catsby (approved):

briancain (approved):

PR #64, opened 2024-05-08
VAULT-26727 - Add support for configuring service configs

codergs:

The PR adds support for configuring profile for Vault Secrets service in addition to setting project and org id. The profile init workflow now will prompt the users to check if they want to configure service profile or not. If yes is chosen, then user will be promoted to select HVS application. 

### How I've tested this PR:
- [x] unit test
- [x] local cli run test 
```
$ hcp profile init
Use the arrow keys to navigate: ↓ ↑ → ←  and / toggles search
? Multiple organizations found. Please select the one you would like to configure.: 
  >test-org
 test-2-org

----- Organization -----
Name:   test-org
ID:     5156c230-562f-42aa-9afe-4582c6ab1e90
```
```
....
✓ Organization with name "test-org" and ID "5156c230-562f-42aa-9afe-4582c6ab1e90" selected
✓ Project with name "test-project" and ID "8e6e60ac-55fd-4906-9ea9-78732432ebdf" selected

Would you like to configure any service related config (y/n)? ? y

Use the arrow keys to navigate: ↓ ↑ → ← 
? Please select the service you would like to configure: 
  ▸ Vault Secrets
```

```
....

✔ Vault Secrets
Use the arrow keys to navigate: ↓ ↑ → ←  and / toggles search
? Multiple apps found. Please select the one you would like to configure.: 
  test-a
  > test-app-2
  test-app-3
  test-app-4

----- Apps -----
Name:   test-app-2
Description This is a test app.
```

```
....
✓ Organization with name "test-org" and ID "5156c230-562f-42aa-9afe-4582c6ab1e90" selected
✓ Project with name "test-project" and ID "8e6e60ac-55fd-4906-9ea9-78732432ebdf" selected

Would you like to configure any service related config (y/n)? y

✔ Vault Secrets
✓ App with name "test-app-2" selected
```

```
$ hcp profile init
$ hcp profile init --vault-secrets
```

### Checklist:


internal/commands/profile/init.go
```diff
@@ -49,6 +56,16 @@ func NewCmdInit(ctx *cmd.Context) *cmd.Command {
 
 			return opts.run()
 		},
+		Flags: cmd.Flags{
+			Local: []*cmd.Flag{
+				{
+					Name:          "vault-secrets",
+					Description:   "Flag to configure current active profile for use with HCP Vault Secrets.",
```
dadgar, line 63: Description shouldn't include that it is a flag. I also think we can drop HCP since it is the `hcp` cli. 

```suggestion
					Description:   "Initialize Vault Secrets configuration.",
```

internal/commands/profile/init.go
```diff
@@ -49,6 +56,16 @@ func NewCmdInit(ctx *cmd.Context) *cmd.Command {
...
+					IsBooleanFlag: true,
+					Value:         flagvalue.Enum([]bool{true, false}, false, &opts.VaultSecrets),
```
dadgar, line 65:
```suggestion
					Value:         flagvalue.Simple(false, &opts.VaultSecrets),
```

internal/commands/profile/init.go
```diff
@@ -62,16 +79,65 @@ type InitOpts struct {
...
+	// Flags
+	VaultSecrets bool
 }
 
 func (i *InitOpts) run() error {
 	if err := i.configureOrgAndProject(); err != nil {
 		return fmt.Errorf("failed configuring organization and project: %w", err)
 	}
 
+	err := i.serviceConfigPrompt()
```
dadgar, line 93: Please add a comment such as

```suggestion
        // If the user hasn't explicitly specified a service to configure, prompt to see if they would like to.
	if err := i.serviceConfigPrompt(); err != nil {
```

internal/commands/profile/init.go
```diff
@@ -62,16 +79,65 @@ type InitOpts struct {
...
+	if i.VaultSecrets {
+		if err := i.configureVaultSecrets(); err != nil {
+			return fmt.Errorf("failed configuring profile for vault secrets: %w", err)
+		}
+	}
+
 	return i.Profile.Write()
 }
 
+func (i *InitOpts) serviceConfigPrompt() error {
```
dadgar, line 107: Add a comment to the function saying what it does.

internal/commands/profile/init.go
```diff
@@ -62,16 +79,65 @@ type InitOpts struct {
...
+	if i.VaultSecrets {
+		return nil
+	}
+
+	if i.IO.CanPrompt() {
```
dadgar, line 112: This isn't needed since the command checks prompting is enabled: https://github.com/hashicorp/hcp/pull/64/files#diff-aae0cdd4820c4e0243986e34d1dd29c5014d2b240bef4b360738cf7e8ff1238cR53

internal/commands/profile/init.go
```diff
@@ -62,16 +79,65 @@ type InitOpts struct {
...
 	}
 
+	err := i.serviceConfigPrompt()
+	if err != nil {
+		return err
+	}
+
+	if i.VaultSecrets {
+		if err := i.configureVaultSecrets(); err != nil {
+			return fmt.Errorf("failed configuring profile for vault secrets: %w", err)
```
dadgar, line 100: Should "vault secrets" be "Vault Secrets"

internal/commands/profile/init.go
```diff
@@ -103,6 +169,62 @@ func (i *InitOpts) configureOrgAndProject() error {
...
+	listAppReq.ProjectID = i.Profile.ProjectID
+	listAppResp, err := i.SecretService.ListApps(listAppReq, nil)
+	if err != nil {
+		return err
+	}
+
+	appCount := len(listAppResp.Payload.Apps)
+
+	if appCount <= 0 {
+		return fmt.Errorf("there are no valid vault secrets apps for your principal. Create one by visiting the HCP Portal (https://portal.cloud.hashicorp.com)")
```
dadgar, line 185: Should this be an error or just an info message printed to `Stderr`? Also should we instruct the user to use `hcp vault-secrets apps create` instead?

internal/commands/profile/property_docs.go
```diff
@@ -50,6 +51,12 @@ func addCoreProperties(b *availablePropertiesBuilder) {
 		and {{ template "mdCodeOrBold" "json" }}.`)
 }
 
+func addVaultSecretsProperties(b *availablePropertiesBuilder) {
+	// vault-secrets
+	b.AddProperty("vault-secrets", "no_color", "If True, color will not be used when printing messages in the terminal.")
```
dadgar, line 56: Copy paste error?

internal/commands/profile/property_docs.go
```diff
@@ -50,6 +51,12 @@ func addCoreProperties(b *availablePropertiesBuilder) {
 		and {{ template "mdCodeOrBold" "json" }}.`)
 }
 
+func addVaultSecretsProperties(b *availablePropertiesBuilder) {
+	// vault-secrets
```
dadgar, line 55: Does the comment add value?

dadgar (changes requested):

internal/commands/profile/init.go
codergs: good idea,  I was following a similar setup to proj, but makes sense to impart more info here.

internal/commands/profile/init.go
codergs: I think why I added is coz current PromptConfirm raises an error, but I was trying to bypass promoting if prompt is disabled.

internal/commands/profile/init.go
```diff
@@ -103,6 +171,67 @@ func (i *InitOpts) configureOrgAndProject() error {
...
+	if err != nil {
+		return err
+	}
+
+	appCount := len(listAppResp.Payload.Apps)
+	if appCount <= 0 {
+		appsCreateDoc := heredoc.New(i.IO, heredoc.WithPreserveNewlines()).Must(`
+No Vault Secrets application found. Create one and set on the active profile by issuing:
+	$ hcp vault-secrets apps create test-app --description="Test app"
+	$ hcp profile set vault-secrets/app_name test-app
```
dadgar, line 189:
```suggestion
No Vault Secrets application found. Create one and set on the active profile by issuing:

	$ hcp vault-secrets apps create test-app --description="Test app"
	$ hcp profile set vault-secrets/app_name test-app
```

.github/CODEOWNERS
```diff
@@ -15,3 +15,4 @@ internal/pkg/waypoint @hashicorp/cloud-waypoint
 
 # Vault Secrets Owned
 *vaultsecrets* @hashicorp/cloud-vault-secrets
+*vault_secrets* @hashicorp/cloud-vault-secrets
```
dadgar, line 18: Can we make the paths explicit?

internal/commands/profile/init.go
```diff
@@ -15,13 +15,19 @@ import (
...
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
+	"github.com/hashicorp/hcp/internal/pkg/flagvalue"
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
 	"github.com/hashicorp/hcp/internal/pkg/iostreams"
 	"github.com/hashicorp/hcp/internal/pkg/profile"
 	"github.com/manifoldco/promptui"
 )
 
+const (
+	ServiceNameVaultSecrets = "Vault Secrets"
```
dadgar, line 28: Does this const need to be exposed?

dadgar (approved):

.github/CODEOWNERS
codergs: I actually prefer it wider at this point as recently in terraform-provider-hcp we had lots of issues with being explicit. We can lock them down later when a merge to main is planned. Hoping this is a nit at this point .

dadgar: It is a nit but it seems like it will just be `internal/commands/vaultsecrets`. The shared directories are already open to all

codergs: I added one under profile too. If you feel strong for being explicit, then I can do that tomorrow.

dadgar: I think it is okay for platform to be code reviewers when the profile is being modified. That is a resource that we will want to have some consistency across services. Don't want a team to potentially use profiles in an unintended way.

codergs: ok, so in that then this entry isn't needed. For the other one, I'll make the other one explicit.

PR #72, opened 2024-05-15
Adds help text after vault-secret app and secret creates

mercedesbh:
### How I've tested this PR:
Ran `make go/build` to test local changes

1. Checkout this branch
2. Run make go/build
3. Create an app and view the help text
4. Create a secret and view the help text

Screenshot 2024-05-15 at 10 24 57 AM

Not going with the below route - see comments below and [this Slack thread]

(https://hashicorp.slack.com/archives/C0630MA5H8B/p1715803590808439?thread_ts=1715721973.724659&cid=C0630MA5H8B) for more details
~~Without app name in help text (Current). We're opting to omit the app name flag from the help command since (1) users will likely have it in their profile, and (2) it may mislead users to believe it is required in the command.~~
Screenshot 2024-05-15 at 11 00 31 AM

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/apps/create.go
+++ b/internal/commands/vaultsecrets/apps/create.go
@@ -104,7 +104,7 @@ func createRun(opts *CreateOpts) error {
 	if opts.Output.GetFormat() == format.Unset {
 		fmt.Fprintf(opts.IO.Err(), "%s Successfully created application with name %q\n", opts.IO.ColorScheme().SuccessIcon(), opts.AppName)
 
-		command := fmt.Sprintf(`$ hcp vault-secrets secrets create <secret name> --app %s --data-file <path to secret>`, opts.AppName)
+		command := "$ hcp vault-secrets secrets create <secret name> --data-file <path to secret>"
 
 		fmt.Fprintln(opts.IO.Err())
 		fmt.Fprintf(opts.IO.Err(), `To create a secret in the app, run:
--- a/internal/commands/vaultsecrets/apps/create.go
+++ b/internal/commands/vaultsecrets/apps/create.go
@@ -104,7 +104,7 @@ func createRun(opts *CreateOpts) error {
 	if opts.Output.GetFormat() == format.Unset {
 		fmt.Fprintf(opts.IO.Err(), "%s Successfully created application with name %q\n", opts.IO.ColorScheme().SuccessIcon(), opts.AppName)
 
-		command := "$ hcp vault-secrets secrets create <secret name> --data-file <path to secret>"
+		command := fmt.Sprintf(`$ hcp vault-secrets secrets create <secret name> --app %s --data-file <path to secret>`, opts.AppName)
 
 		fmt.Fprintln(opts.IO.Err())
 		fmt.Fprintf(opts.IO.Err(), `To create a secret in the app, run:
--- a/internal/commands/vaultsecrets/apps/create.go
+++ b/internal/commands/vaultsecrets/apps/create.go
@@ -44,9 +44,6 @@ func NewCmdCreate(ctx *cmd.Context, runF func(*CreateOpts) error) *cmd.Command {
 		ShortHelp: "Create a new Vault Secrets application.",
 		LongHelp: heredoc.New(ctx.IO).Must(`
 		The {{ template "mdCodeOrBold" "hcp vault-secrets apps create" }} command creates a new Vault Secrets application.
-
-		Once an application is created, secrets lifecycle can be managed using the
-		{{ template "mdCodeOrBold" "hcp vault-secrets secrets" }} command group.
 		`),
 		Examples: []cmd.Example{
 			{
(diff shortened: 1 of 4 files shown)


mercedesbh:
These help commands won't display if the user specifies a format. Still thinking of a way to have it output in all scenarios (and deciding if we even want it to display when the user asks for a specific format).

codergs:
> These help commands won't display if the user specifies a format. Still thinking of a way to have it output in all scenarios (and deciding if we even want it to display when the user asks for a specific format).

Personally, I am happy with just showing help only without format specification. It gets the work done, and also doesn't complicate decision making.

internal/commands/vaultsecrets/apps/create.go
```diff
@@ -106,6 +103,13 @@ func createRun(opts *CreateOpts) error {
 
 	if opts.Output.GetFormat() == format.Unset {
 		fmt.Fprintf(opts.IO.Err(), "%s Successfully created application with name %q\n", opts.IO.ColorScheme().SuccessIcon(), opts.AppName)
+
+		command := "$ hcp vault-secrets secrets create <secret name> --data-file <path to secret>"
+
+		fmt.Fprintln(opts.IO.Err())
```
codergs, line 109: is it easy to use heredoc here?

dadgar:
@mercedesbh @codergs Curious on this statement:

> We're opting to omit the app name flag from the help command since (1) users will likely have it in their profile

Why would that be true? The user just created the application? If you want that behavior maybe have a flag? Here is an example: https://github.com/hashicorp/hcp/blob/main/internal/commands/projects/create.go#L68

codergs:
> @mercedesbh @codergs Curious on this statement:
> 
> > We're opting to omit the app name flag from the help command since (1) users will likely have it in their profile
> 
> Why would that be true? The user just created the application? If you want that behavior maybe have a flag? Here is an example: https://github.com/hashicorp/hcp/blob/main/internal/commands/projects/create.go#L68
> 
> I would also say, even if it is set in the profile, and the flag is provided, things will still work as expected. Providing the app flag at least guarantees that the secret is being created in the app they just created.

I can't see that comment, but I don't necessarily thing flag override is needed here as by showing that we implicitly giving an impression flag may or may not be required. The -h command should already cover that use case.

internal/commands/vaultsecrets/apps/create.go
mercedesbh: Didn't try with heredoc, but figured this approach is more consistent other implementations in the codebase ([here](https://github.com/hashicorp/hcp/blob/9d3adac44cf8431bd5ea9e53b3fa53fdf9548261/internal/commands/iam/workloadidentityproviders/create_aws.go#L187-L196) as an example).

mercedesbh:
> > @mercedesbh @codergs Curious on this statement:
> > > We're opting to omit the app name flag from the help command since (1) users will likely have it in their profile
> > 
> > 
> > Why would that be true? The user just created the application? If you want that behavior maybe have a flag? Here is an example: https://github.com/hashicorp/hcp/blob/main/internal/commands/projects/create.go#L68
> > I would also say, even if it is set in the profile, and the flag is provided, things will still work as expected. Providing the app flag at least guarantees that the secret is being created in the app they just created.
> 
> I can't see that comment, but I don't necessarily think the app name flag override is needed here as by showing that we implicitly giving an impression flag may or may not be required. The -h command should already cover that use case.

@codergs and I had chatted about it a bit in Slack, but I think @dadgar's point about the app being newly created (hence not being in the profile yet), is a good point. And specifying the app name even if it is in the profile is harmless.

internal/commands/vaultsecrets/apps/create.go
codergs: Sounds good to me if an existing usage pattern in the repo. I wasn't sure, but feel free to close this out. Thanks!

codergs:
> > > @mercedesbh @codergs Curious on this statement:
> > > > We're opting to omit the app name flag from the help command since (1) users will likely have it in their profile
> > > 
> > > 
> > > Why would that be true? The user just created the application? If you want that behavior maybe have a flag? Here is an example: https://github.com/hashicorp/hcp/blob/main/internal/commands/projects/create.go#L68
> > > I would also say, even if it is set in the profile, and the flag is provided, things will still work as expected. Providing the app flag at least guarantees that the secret is being created in the app they just created.
> > 
> > 
> > I can't see that comment, but I don't necessarily think the app name flag override is needed here as by showing that we implicitly giving an impression flag may or may not be required. The -h command should already cover that use case.
> 
> @codergs and I had chatted about it a bit in Slack, but I think @dadgar's point about the app being newly created (hence not being in the profile yet), is a good point. And specifying the app name even if it is in the profile is harmless. Not sure if we should be setting the newly created app as default for them.

Ahh coz profile won't be set for it. Actually, nvm then. yeah, seems fine with me. For secrets create maybe we don't need as app name validation are in place.

codergs (approved):
LGTM.

PR #77, opened 2024-05-15
Fixup: Rename -application-name to just -app for consistency

briancain:

Other HCP commands use -app for the app name, and it's much less typing than -application-name. This simplifies this flag value.

### How I've tested this PR:

Built and validated the flag name changed, updated the tests and saw they passed.

Same ☝🏻 

### Checklist:

### DIFF
--- a/internal/commands/waypoint/add-ons/create.go
+++ b/internal/commands/waypoint/add-ons/create.go
@@ -59,7 +59,7 @@ $ hcp waypoint add-ons create -n=my-addon -a=my-application -d=my-addon-definiti
 				},
 				{
 					Name:         "app",
-					DisplayValue: "APP-NAME",
+					DisplayValue: "NAME",
 					Description:  "The name of the application to which the add-on will be added.",
 					Value:        flagvalue.Simple("", &opts.ApplicationName),
 					Required:     true,
--- a/internal/commands/waypoint/add-ons/create.go
+++ b/internal/commands/waypoint/add-ons/create.go
@@ -58,8 +58,8 @@ $ hcp waypoint add-ons create -n=my-addon -a=my-application -d=my-addon-definiti
 					Required:     true,
 				},
 				{
-					Name:         "application-name",
-					DisplayValue: "NAME",
+					Name:         "app",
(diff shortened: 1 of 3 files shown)


internal/commands/waypoint/add-ons/create.go
```diff
@@ -58,8 +58,8 @@ $ hcp waypoint add-ons create -n=my-addon -a=my-application -d=my-addon-definiti
 					Required:     true,
 				},
 				{
-					Name:         "application-name",
-					DisplayValue: "NAME",
+					Name:         "app",
+					DisplayValue: "APP-NAME",
```
codergs, line 62: nit: vaultsecrets were advised to keep it at NAME only

internal/commands/waypoint/add-ons/create.go
```diff
@@ -58,8 +58,8 @@ $ hcp waypoint add-ons create -n=my-addon -a=my-application -d=my-addon-definiti
 					Required:     true,
 				},
 				{
-					Name:         "application-name",
-					DisplayValue: "NAME",
+					Name:         "app",
```
codergs, line 61: I wonder since this is a breaking change, how about we add both into the mix for some period of time, and then let users drift away. Not sure if it is possible, but when specified both could update the same underlying var.

briancain: I don't feel like that's worth the maintenance effort. Maybe if this CLI had been around for years sure, but at this point it's brand new (not to mention Waypoint is still in beta) so I'd rather make the breaking change than maintain two flags for backwards compatibility right now. I do appreciate the suggestion though!

internal/commands/waypoint/add-ons/create.go
briancain: Got it, fixed in [fbe2710](https://github.com/hashicorp/hcp/pull/77/commits/fbe271054a7e05d97b1168d5e6ecb166d9ccc9f3)

codergs (approved):
LGTM. Please fill the description section if possible 👍

paladin-devops (approved):
The add-on-appening! ✅

PR #79, opened 2024-05-16
VAULT-26824 - Add hcp vault-secrets secrets read 

codergs:

This PR adds the `hcp vault-secrets secrets read` command to the `hcp` CLI. 

### How I've tested this PR:
- [x] locally ran the binary to read a HVS secret
```
>  make go/build && ./bin/hcp vault-secrets secrets read -h       
USAGE
  hcp vault-secrets secrets read NAME [Optional Flags]

DESCRIPTION
  The hcp vault-secrets secrets read command reads a static secret from the Vault
  Secrets application.

EXAMPLES
  Read secret's metadata:
  
  $ hcp vault-secrets secret read "test_secret"
  
  Read plaintext secret:
  
  $ hcp vault-secrets secret read "test_secret" --plaintext

POSITIONAL ARGUMENTS
  NAME
    The name of the secret to read.

FLAGS
  -p, --plaintext
    Display the secret's value in plaintext.

INHERITED FLAGS
  -a, --app=NAME
    The name of the Vault Secrets application. If not specified, the value from the
    active profile will be used.

GLOBAL FLAGS
  --debug, --format=FORMAT, --profile=NAME, --project=ID, --quiet
  
  For more global flag details, run $ hcp --help

> ./bin/hcp vault-secrets secrets read test_100 --app test-gs
Secret Name  Latest Version  Created At                
test_100     1               2024-05-15T15:14:36.457Z  
....

> ./bin/hcp vault-secrets secrets read temp -p --app test-gs
Secret Name:    temp
Latest Version: 13
Created At:     2024-05-17T14:47:24.104Z
Value:          temp

> make go/build && ./bin/hcp vault-secrets secrets read temp --app test-gs 
Secret Name:    temp
Latest Version: 13
Created At:     2024-05-17T14:47:24.104Z
```

- [x] Unit test
 
```
$ make go/build 
$ ./bin/hcp auth login 
$ ./bin/hcp profile init --vault-secrets
$ ./bin/hcp vault-secrets secrets read
```

### Checklist:


internal/commands/vaultsecrets/secrets/read.go
```diff
@@ -0,0 +1,145 @@
...
+				{
+					Name:          "NAME",
+					Documentation: "The name of the secret to read.",
+				},
+			},
+		},
+		Flags: cmd.Flags{
+			Local: []*cmd.Flag{
+				{
+					Name:          "plaintext",
```
dadgar, line 61: Would an alias be good here since it will be common

internal/commands/vaultsecrets/secrets/read.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("failed to read the secret %q: %w", opts.SecretName, err)
+	}
+	return opts.Output.Display(newDisplayer(true).Secrets(resp.Payload.Secret))
+}
+
+func runOpenAppSecret(opts *ReadOpts) error {
+	// Table format doesn't handle large secret values well. So until that is fixed,
+	// we should not allow table format for this command.
+	if opts.Output.GetFormat() == format.Table {
+		return fmt.Errorf("table format is not supported for this command")
```
dadgar, line 117: Let's not do this. If the table output isn't well formatted, the user can specify an alternative format.

internal/commands/vaultsecrets/secrets/read.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	}
+
+	d := newDisplayer(true).OpenAppSecrets(resp.Payload.Secret)
+	d.AddFields([]format.Field{
+		{
+			Name:        "Value",
+			ValueFormat: "{{.Version.Value}}",
+		},
+	})
```
dadgar, line 142: Hm this doesn't feel right to me. There is no way for a user to specify format.Unset. So there is no way to get to this behavior if the profile has a default output format set.

internal/commands/vaultsecrets/secrets/read.go
codergs: Trying to understand - we don't handle this well today, so why put it out there?

internal/commands/vaultsecrets/secrets/read.go
codergs: hmmn.. can format be set on profile as well? in my usual runs, it is hitting that block - so without user specifying it, we hit that in the base case. what am i missing?

codergs: Ok, there is  `"core/output_format": {"pretty", "table", "json"}`. I think we should discuss this 1-1.

internal/commands/vaultsecrets/secrets/read.go
codergs: After discussing, we think it ok to add it as for shorter creds this won't be a problem.

internal/commands/vaultsecrets/secrets/read.go
codergs: cc: @mercedesbh

internal/commands/vaultsecrets/secrets/read.go
```diff
@@ -0,0 +1,135 @@
...
+	req.LocationProjectID = opts.Profile.ProjectID
+	req.AppName = opts.AppName
+	req.SecretName = opts.SecretName
+
+	resp, err := opts.Client.OpenAppSecret(req, nil)
+	if err != nil {
+		return fmt.Errorf("failed to read the secret %q: %w", opts.SecretName, err)
+	}
+
+	d := newDisplayer(true).OpenAppSecrets(resp.Payload.Secret).SetDefaultFormat(format.Pretty)
```
dadgar, line 126: Consider having both the open and read have the same default format.

dadgar (approved):

internal/commands/vaultsecrets/secrets/read.go
```diff
@@ -123,7 +123,7 @@ func runOpenAppSecret(opts *ReadOpts) error {
 		return fmt.Errorf("failed to read the secret %q: %w", opts.SecretName, err)
 	}
 
-	d := newDisplayer(true).OpenAppSecrets(resp.Payload.Secret).SetDefaultFormat(format.Pretty)
```
dadgar: Just to clarify I was actually thinking both should use pretty.

Issue #68, opened 2024-05-13
iostream ReadSecret should honor user interrupts


codergs:
The `iostreams` packaged system ReadSecret doesn't honor user interrupts.

**Observed:** 
While working on [PR](https://github.com/hashicorp/hcp/pull/65), it was seen that once the user is prompted to provide plaintext via secret value, ctrl +C doesn't cancel the command. The only way to abort is to try something like ctrl +Z and then kill the process. Obviously, I can hit enter and get around this as well. 

```
$ go env | grep -e GOARCH -e GOHOSTOS -e GOOS -e GOVERSION
GOARCH='amd64'
GOHOSTOS='darwin'
GOOS='darwin'
GOVERSION='go1.22.2'
```
**Reproduce:**
Build the CLI using the mentioned [PR](https://github.com/hashicorp/hcp/pull/65), and issue 

```
$ ./bin/hcp auth login
$ ./bin/hcp profile init --vault-secrets
$ ./bin/hcp vault-secrets secrets create test       
Please enter the plaintext secret: 
```
Now try hitting `ctrl + C`.

**Expected Behavior or equivalent:**
```
$ ./bin/hcp vault-secrets secrets create test       
Please enter the plaintext secret: 
```
Pressing ^Ctrl +C should terminate the run gracefully.

dadgar:
Closed by #83

PR #83, opened 2024-05-17
HCPCP-1620: ReadSecret: Respect Ctrl-C Interrupt

jasonpilz:
* Relates to https://github.com/hashicorp/hcp/issues/68 and HCPCP-1620
* Updates the `ReadSecret` method to listen for the canceled context signaled by `os.Interrupt`, logs the cancel cause and exits the program.

### How I've tested this PR:
* NOTE: This bug was surfaced with vault secrets having the option for STDIN password. This has since be removed but is being addressed for future use. Due to this, it was tested on a historical git commit. 

Screenshot 2024-05-20 at 9 52 52 AM

Tested with Iterm and Bash:
```
$ git checkout test-jason

$ ./bin/hcp auth login
$ ./bin/hcp profile init --profile=vault-secrets
$ ./bin/hcp profile set vault-secrets/app_name <app_name>
$ ./bin/hcp vault-secrets secrets create test

^c
```

### Checklist:

### DIFF
--- a/internal/pkg/iostreams/io.go
+++ b/internal/pkg/iostreams/io.go
@@ -13,9 +13,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/signal"
 	"strings"
-	"syscall"
 
 	"github.com/muesli/termenv"
 	"golang.org/x/term"
@@ -177,20 +175,13 @@ func (s *system) ReadSecret() ([]byte, error) {
 	}
 	errorChannel := make(chan Buffer, 1)
 
-	// SIGINT and SIGTERM restore the terminal, otherwise the no-echo mode would remain intact
-	interruptChannel := make(chan os.Signal, 1)
-	signal.Notify(interruptChannel, syscall.SIGINT, syscall.SIGTERM)
-	defer func() {
-		signal.Stop(interruptChannel)
-		close(interruptChannel)
-	}()
+	// Cancelled context restores the terminal, otherwise the no-echo mode would remain intact
 	go func() {
-		for range interruptChannel {
-			if oldState != nil {
-				_ = term.Restore(fd, oldState)
-			}
-			errorChannel <- Buffer{Buffer: make([]byte, 0), Error: ErrInterrupt}
+		<-s.ctx.Done()
+		if oldState != nil {
+			_ = term.Restore(fd, oldState)
 		}
+		errorChannel <- Buffer{Buffer: make([]byte, 0), Error: ErrInterrupt}
 	}()
 
 	go func() {
--- a/internal/pkg/iostreams/io.go
+++ b/internal/pkg/iostreams/io.go
@@ -12,7 +12,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/muesli/termenv"
 	"golang.org/x/term"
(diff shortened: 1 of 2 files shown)


internal/pkg/iostreams/io.go
```diff
@@ -156,6 +156,13 @@ func (s *system) ReadSecret() ([]byte, error) {
 		return nil, fmt.Errorf("prompting is disabled")
 	}
 
+	go func() {
```
dadgar, line 159: I tested this and unfortunately it breaks the terminal state. You can't use up/down errors to go through history, command isn't shown, etc.

Through some spelunking, you need to call RestoreState: https://github.com/golang/go/issues/31180#issuecomment-1133854964

I did run across this implementation which works really well: https://github.com/grisu48/common/blob/main/pkg/password/password_supported.go#L17

I think we should bring that implementation to this package. I like that it returns the error to the caller as well

dadgar (changes requested):

dadgar:
@jasonpilz Thanks for taking a look at this issue ❤️

internal/pkg/iostreams/io.go
jasonpilz: Great find as this was hidden when using Zsh.
Have imported state handling as suggested. 🙇🏻

internal/pkg/iostreams/io.go
```diff
@@ -156,7 +161,46 @@ func (s *system) ReadSecret() ([]byte, error) {
...
 
-	return term.ReadPassword(int(s.in.Fd()))
+	fd := int(s.in.Fd())
+
+	// Store and restore the terminal status on interruptions to
+	// avoid that the terminal remains in the password state
+	// This is necessary as for https://github.com/golang/go/issues/31180
+	oldState, err := term.GetState(fd)
+	if err != nil {
+		return make([]byte, 0), err
```
dadgar, line 171:
```suggestion
		return nil, err
```

Nil should be fine on error

internal/pkg/iostreams/io.go
```diff
@@ -156,7 +161,46 @@ func (s *system) ReadSecret() ([]byte, error) {
...
+
+	type Buffer struct {
+		Buffer []byte
+		Error  error
+	}
+	errorChannel := make(chan Buffer, 1)
+
+	// SIGINT and SIGTERM restore the terminal, otherwise the no-echo mode would remain intact
+	interruptChannel := make(chan os.Signal, 1)
+	signal.Notify(interruptChannel, syscall.SIGINT, syscall.SIGTERM)
```
dadgar, line 182: So the context will already be cancelled from those signals: https://github.com/hashicorp/hcp/blob/main/main.go#L42

So instead, just listen to the context being cancelled.

dadgar (changes requested):

internal/pkg/iostreams/io.go
jasonpilz: Thank you. Copy 🍝 got the better of me

internal/pkg/iostreams/io.go
```diff
@@ -156,7 +159,39 @@ func (s *system) ReadSecret() ([]byte, error) {
...
+	}
+	errorChannel := make(chan Buffer, 1)
+
+	// Cancelled context restores the terminal, otherwise the no-echo mode would remain intact
+	go func() {
+		<-s.ctx.Done()
+		if oldState != nil {
+			_ = term.Restore(fd, oldState)
+		}
+		errorChannel <- Buffer{Buffer: make([]byte, 0), Error: ErrInterrupt}
```
dadgar, line 184:
```suggestion
	errorChannel := make(chan Buffer, 1)
	doneChannel := make(chan struct{})
	defer close(doneChannel)

	// Cancelled context restores the terminal, otherwise the no-echo mode would remain intact
	go func() {
		select {	
		case <-doneChannel:
			return
		case <-s.ctx.Done():
		}
		
		if oldState != nil {
			_ = term.Restore(fd, oldState)
		}
		errorChannel <- Buffer{Buffer: make([]byte, 0), Error: context.Cause(s.ctx)}
```

So these are nits, but suggestion two changes:

1. Add a doneChannel that causes the goroutine to exit. This is a nit, since it is a CLI and isn't expected to be long running anyways, but as it is, this leaks a goroutine until the CLI exits.
2. Lets not add the ErrInterrupt but instead use context.Cause like you had it before.

internal/pkg/iostreams/io.go
```diff
@@ -18,6 +19,8 @@ import (
 	"golang.org/x/term"
 )
 
+var ErrInterrupt = errors.New("interrupted")
```
jasonpilz, line 22:
```suggestion
```

internal/pkg/iostreams/io.go
jasonpilz: ❤️ it.

dadgar (approved):
Great work!

PR #85, opened 2024-05-17
Add autocomplete for the apps/secrets -> read/delete command

codergs:

The PR makes follwoing changes 

1. adds a `s` alias for the `secrets` subcommand
2. adds auto complete functionality for `--app` flag
3. adds auto complete fucntionality for `app` and `secret name` args on apps/secrets read/delete paths.
4. Removes any references to Beta
5. Added an autocomplete for --data-file flag under `hcp vs s create` similar to [this](https://github.com/hashicorp/hcp/blob/b59b5ece6ed275fdec5f35e1046b17d601503411/internal/commands/auth/login.go#L110)

### How I've tested this PR:
-[x] built locally and exercised tab to see autocomplete pick up/show the values

```
$ make go/build
$ ./bin/hcp v...keep entering tab until you craft the command you need
```

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/secrets/delete.go
+++ b/internal/commands/vaultsecrets/secrets/delete.go
@@ -54,7 +54,7 @@ func NewCmdDelete(ctx *cmd.Context, runF func(*DeleteOpts) error) *cmd.Command {
 					Documentation: "The name of the secret to create.",
 				},
 			},
-			Autocomplete: helper.PredictSecretName(ctx, ctx.Profile.OrganizationID, ctx.Profile.ProjectID, appname.Get(), opts.PreviewClient),
+			Autocomplete: helper.PredictSecretName(ctx, ctx.Profile.OrganizationID, ctx.Profile.ProjectID, opts.PreviewClient),
 		},
 		RunF: func(c *cmd.Command, args []string) error {
 			opts.AppName = appname.Get()
--- a/internal/commands/vaultsecrets/secrets/delete.go
+++ b/internal/commands/vaultsecrets/secrets/delete.go
@@ -10,6 +10,7 @@ import (
 	preview_secret_service "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/client/secret_service"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/stable/2023-06-13/client/secret_service"
 	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/secrets/appname"
+	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/secrets/helper"
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
 	"github.com/hashicorp/hcp/internal/pkg/format"
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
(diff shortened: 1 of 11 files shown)


internal/commands/vaultsecrets/secrets/delete.go
```diff
@@ -53,6 +54,7 @@ func NewCmdDelete(ctx *cmd.Context, runF func(*DeleteOpts) error) *cmd.Command {
 					Documentation: "The name of the secret to create.",
 				},
 			},
+			Autocomplete: helper.PredictSecretName(ctx, ctx.Profile.OrganizationID, ctx.Profile.ProjectID, appName, opts.PreviewClient),
```
codergs, line 57: arg level autocomplete isn't working in my terminal. will check again on Monday.

internal/commands/vaultsecrets/secrets/helper/helper.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// PredictAppName returns a predict function for application names.
+func PredictSecretName(ctx *cmd.Context, orgID, projectID string, client preview_secret_service.ClientService) complete.PredictFunc {
+	return func(args complete.Args) []string {
+		if len(args.Completed) > 1 {
+			return nil
+		}
+
+		appName := appname.Get()
```
codergs, line 24: `appname.Get()` is yielding empty as I have a feeling at this time the `vault-secrets secrets` command's `PersistentPreRun` has not run (maybe ?). checking with @dadgar

dadgar (approved):

internal/commands/vaultsecrets/secrets/delete.go
```diff
@@ -64,6 +65,7 @@ func NewCmdDelete(ctx *cmd.Context, runF func(*DeleteOpts) error) *cmd.Command {
 			return deleteRun(opts)
 		},
 	}
+	cmd.Args.Autocomplete = helper.PredictSecretName(ctx, cmd, opts.PreviewClient)
```
codergs, line 68: Thanks @dadgar. What changes when defined like this vs inline before?

dadgar: I needed access to the `cmd` variable, so it couldn't be defined in line.

codergs: I see the cmd injection logic, but I didn't need it as such when using the profile set app name. I am trying to understand how appname is populated now vs if when we were inlining autocomplete.

internal/commands/vaultsecrets/secrets/secrets.go
```diff
@@ -34,6 +37,13 @@ func NewCmdSecrets(ctx *cmd.Context) *cmd.Command {
 		},
 	}
 
+	// Autocomplete the persistent flag.
+	for _, f := range cmd.Flags.Persistent {
```
codergs, line 41: Similar question here. This was working with inline autocomplete. Can't we inline this as well?

dadgar: Same answer as above. You need to parse the flags in the autocomplete function and thus you need access to the command.

internal/commands/vaultsecrets/secrets/delete.go
codergs: Gotcha. I think my original question is more on the lines that `AutoComplete` is always going to need these flag or ability to parse these flags. If we need to set this up this way for all commands, then inlining use case is really not there. Maybe we have need a setter construct and autocomplete becomes internal detail. This gets job done, but the whole app name lookup under persistence flag also needs to hide internally in provided frameworks (on the other comment). For now, this is ok.

PR #91, opened 2024-05-22
Detect new version

dadgar:
Detect new versions and prompt the user to upgrade.

### How I've tested this PR:
Unit tests, and changed the version to be < 0.1.0 and saw the output. Also tested deleting / corrupting the version check state file.

Code review, test on your machine by pulling the branch and modifying `version/VERSION` to be a lower version.

### Output of affected commands:
Screenshot 2024-05-21 at 8 02 15 PM

### Checklist:


internal/pkg/versioncheck/detect.go
```diff
@@ -0,0 +1,226 @@
...
+	if _, err := os.Stat(dir); err != nil {
+		// If the directory doesn't exist, create it.
+		if errors.Is(err, fs.ErrNotExist) {
+			if err := os.MkdirAll(dir, 0766); err != nil {
+				return nil, fmt.Errorf("failed to created HCP config directory %q: %w", dir, err)
+			}
+		} else {
+			return nil, fmt.Errorf("failed to check if HCP config directory exists: %w", err)
+		}
+	}
```
tobias-hashicorp, line 76: This will create the folders for the state file even if the version will not get checked and the state file hence not used. Should we move this into the `Check` method?

internal/pkg/versioncheck/state.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/hashicorp/hcp/internal/pkg/api/releasesapi/models"
+)
+
+// versionCheckState captures when the last check for a new version of the CLI
+// occurred.
+type versionCheckState struct {
+	CheckedAt time.Time `json:"checked_at"`
+
+	// latestRelease is the latest version of the CLI.
+	latestRelease *models.ProductReleaseResponseV1
```
tobias-hashicorp, line 18: Should we also store the result of the last check so we can keep displaying the message even if we do not perform another check?

main.go
```diff
@@ -56,6 +63,20 @@ func realMain() int {
 		}
 	}()
 
+	// Create the version checker
+	checker, err := versioncheck.New(io, versionCheckStatePath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to configure version checker: %v\n", err)
+		return 1
```
tobias-hashicorp, line 70: Should this just be a warning and not cause the application to exit? 

Especially if we expect `New` to create the folders it would otherwise prevent the CLI from being used without write access.

main.go
```diff
@@ -117,6 +137,11 @@ func realMain() int {
 		fmt.Fprintf(io.Err(), "Error executing hcp: %s\n", err.Error())
 	}
 
+	// Cancel the update check if it hasn't finished already
+	if checker.UpdateToDisplay() && !isAutocomplete() {
```
tobias-hashicorp, line 141: Would it make sense to remove the `UpdateToDisplay` function and just rely on `Display` to be a noop if there is nothing to display?

tobias-hashicorp (approved):

main.go
```diff
@@ -56,6 +63,20 @@ func realMain() int {
...
 
+	// Create the version checker
+	checker, err := versioncheck.New(io, versionCheckStatePath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to configure version checker: %v\n", err)
+		return 1
+	}
+
+	// Start checking for a new version as soon as possible
+	go func() {
```
codergs, line 74: Since we plan to only show the new version prompt once per 24 hours, is there a need to put this on a go routine?

internal/pkg/versioncheck/state.go
dadgar: The reason to not prompt on every run wasn't to reduce load on releases but rather to not annoy the operator. Make sense?

internal/pkg/versioncheck/detect.go
dadgar: I could go either way. I put it here because the New() is setting up the checker to successfully save its state. Gives an earlier place to handle an invalid state path.

main.go
dadgar: The reason it is in a goroutine is so that the command the user wants to run and the version checking can run in parallel. This means the user ideally doesn't experience any additional delay because we are doing a version check.

codergs: My concern was more on the lines that a user hits the end while go routine isn't finished. So prompt for version check never happens as it is a slow path.

dadgar: That is the intention. If the command finishes before the version check finishes, you just cancel the version check.

PR #92, opened 2024-05-22
VAULT-27295 implement vault-secrets run

mercedesbh:
This PR adds the `hcp vault-secrets run` (code [ported over from vlt](https://github.com/hashicorp/vlt/blob/main/internal/command/run.go)) command to the hcp CLI for ticket https://hashicorp.atlassian.net/browse/VAULT-27295

### How I've tested this PR:
Ran make go/build to manually test local changes, and added unit tests.

1. Checkout this branch
2. `Run make go/build`
3. Create an app using ./bin/hcp vault-secrets apps create {YOUR_APP_NAME} --description "original description"
4. Add secrets to the app using `./bin/hcp vault-secrets secrets create {YOUR_SECRET_NAME} --data-file {PATH_TO_SECRET}`
5. Run a command using `./bin/hcp vault-secrets run "env"`

```
❯ ./bin/hcp vault-secrets run "go run /Users/mercedeshall/Desktop/billie.go"                                                                                                                       ─╯
Vault Secrets application name must set.

Set the app name using the --app flag or set the app name on your active profile one of the following commands:

$ hcp profile set vault-secrets/app <app_name>
$ hcp profile init --vault-secrets

❯ ./bin/hcp profile set vault-secrets/app company-card                                                                                                                                                       ─╯
✓ Property "vault-secrets/app" updated

❯ ./bin/hcp vault-secrets run "go run /Users/mercedeshall/Desktop/billie.go"                                                                                                                                 ─╯
Simple test app
Vaule of vs as env var for EXPIRATION:  October 2024

❯ ./bin/hcp vault-secrets run -- go run /Users/mercedeshall/Desktop/billie.go                                                                                                                                ─╯
Simple test app
Vaule of vs as env var for EXPIRATION:  October 2024

❯ ./bin/hcp vault-secrets run --help                                                                                                                                                                         ─╯
USAGE
  hcp vault-secrets run COMMAND [Optional Flags]

DESCRIPTION
  The hcp vault-secrets run command lets you run  the provided command as a child
  process while injecting all of the app's secrets as environment variables, with
  all secret names converted to upper-case. The stdout and stderr from the child
  process are forwarded to the top level hcp vault-secrets run command.

EXAMPLES
  Inject secrets as environment variables:
  
  $ hcp vault-secrets run "env"

POSITIONAL ARGUMENTS
  COMMAND
    Defines the invocation of the child process to inject secrets to.

FLAGS
  --app=NAME
    The application you want to pull all secrets from.

GLOBAL FLAGS
  --debug, --format=FORMAT, --profile=NAME, --project=ID, --quiet
  
  For more global flag details, run $ hcp --help
```

### Checklist:


mercedesbh:
Sorting through some things on the rebase...please hold off on review

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+				{
+					Name:         "command",
+					DisplayValue: "COMMAND",
+					Description:  "Defines the invocation of the child process to inject secrets to.",
+					Value:        flagvalue.Simple("", &opts.Command),
+					Required:     true,
+				},
+				{
+					Name:         "app",
+					DisplayValue: "APP",
```
codergs, line 78: this should be NAME.

codergs: This was from @dadgar - we need to support autocomplete for app name here. Kindly, search for AutoComplete under vaultsecrets folder and get an idea on how to use that.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	// get existing environment variables detectable by runtime
+	result := os.Environ()
+
+	for _, secret := range res.Payload.Secrets {
+		// we need to append results in case of duplicates we want secrets to override
+		// only supporting static secrets
+		if secret.StaticVersion != nil {
+			result = append(result, strings.ToUpper(secret.Name)+"="+secret.StaticVersion.Value)
```
codergs, line 147: nit: use fmt.Sprintf

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/hashicorp/hcp/internal/pkg/profile"
+)
+
+type RunOpts struct {
+	Ctx     context.Context
+	Profile *profile.Profile
+	IO      iostreams.IOStreams
+	Output  *format.Outputter
+
+	App           string
```
codergs, line 33: nit: AppName to be consistent with other sub-commands.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+	cmd := pieces[0]
+	var args []string
+	if len(pieces) > 1 {
+		args = pieces[1:]
+	}
+
+	input := &child.NewInput{
+		Stdin:       os.Stdin,
+		Stdout:      os.Stdout,
+		Stderr:      os.Stderr,
```
codergs, line 165: @dadgar do we need to use ctx.IO etc. here or since this is a separate child process we are ok with this setup?

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+		Output:        ctx.Output,
+		PreviewClient: preview_secret_service.New(ctx.HCP, nil),
+		Client:        secret_service.New(ctx.HCP, nil),
+	}
+
+	cmd := &cmd.Command{
+		Name:      "run",
+		ShortHelp: "Run a process with secrets from a Vault Secrets app.",
+		LongHelp: heredoc.New(ctx.IO).Must(`
+		The {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command lets you
```
dadgar, line 53: Should this be `hcp vault-secrets apps run` given you are running a command for all the secrets of an app? This feels like a sub-action of apps?

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+		run the provided command as a child process while injecting
+        all of the app's secrets as environment variables, with all secret names
+        converted to upper-case. The stdout/stderr from the child
+        process are forwarded to the top level 'hcp vault-secrets run' command.
```
dadgar, line 57:
```suggestion
        process are forwarded to the top level {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command.
```

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+	if err := childProcess.Start(); err != nil {
+		return fmt.Errorf("failed to run with secrets in app %q: %w", opts.App, err)
+
+	}
+
+	sigC := make(chan os.Signal, 1)
+	signal.Notify(sigC, os.Interrupt, syscall.SIGTERM)
+	for {
+		select {
+		case <-sigC:
```
dadgar, line 120: Use the opts.Ctx.Done() channel. It will already close if there is a signal.

internal/commands/vaultsecrets/run/run.go
dadgar: I am okay making an exception in this case since we are starting a new process. It will be a bit more performant to pass the file descriptors directly.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+
+func setupChildProcess(command string, envVars []string) (*child.Child, error) {
+	pieces := strings.Split(command, " ")
+	cmd := pieces[0]
+	var args []string
+	if len(pieces) > 1 {
+		args = pieces[1:]
+	}
+
+	input := &child.NewInput{
```
dadgar, line 162: Why are we using consul-template here? That is quite a big dependency to pull in.

Can we use: https://pkg.go.dev/os/exec#CommandContext instead?

dadgar (changes requested):

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,176 @@
...
+		run the provided command as a child process while injecting
+        all of the app's secrets as environment variables, with all secret names
+        converted to upper-case. The stdout/stderr from the child
+        process are forwarded to the top level 'hcp vault-secrets run' command.
+		`),
+		Examples: []cmd.Example{
+			{
+				Preamble: `Inject secrets as an environment variable:`,
+				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
+				$ hcp vault-secrets run --command="env"
```
dadgar, line 63: Should command be an arg instead of a flag?

```
hcp vault-secrets run env
hcp vault-secrets run my-app --config foo
```

internal/commands/vaultsecrets/run/run.go
codergs: I think we should stick with top level for now as we don't know what enhancements are planned in future. Moving it under apps also locks us to one app - a design constraint we don't necessarily have to hold. We should run this through product if we don't anticipate anything more in future.

internal/commands/vaultsecrets/run/run.go
mercedesbh: Ported the run command code over as is, but def a fair point. Moved to using CommandContext

internal/commands/vaultsecrets/run/run.go
mercedesbh: `command` is now an arg

internal/commands/vaultsecrets/run/run.go
mercedesbh: Chatted on this point with the HVS tech lead and PM as well. Sounds like folks are in favor of keeping it top-level for now.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+
+			opts.AppName = appname.Get()
+
+			if runF != nil {
+				return runF(opts)
+			}
+			return runRun(opts)
+		},
+	}
+	cmd.Args.Autocomplete = helper.PredictAppName(ctx, cmd, preview_secret_service.New(ctx.HCP, nil))
```
codergs, line 96: can use opts.PreviewClient

codergs: cmd.Args.Autocomplete is for COMMAND arg, but auto complete is for app flag name. Is this correct usage?

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+					Value:        appname.Flag(),
+				},
+			},
+		},
+		PersistentPreRun: func(c *cmd.Command, args []string) error {
+			return appname.Require(ctx)
+		},
+		RunF: func(c *cmd.Command, args []string) error {
+			opts.Command = args[0]
+
```
codergs, line 87: nit: remove new line

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+		run the provided command as a child process while injecting
+        all of the app's secrets as environment variables, with all secret names
+        converted to upper-case. The stdout/stderr from the child
+        process are forwarded to the top level {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command.
+		`),
+		Examples: []cmd.Example{
+			{
+				Preamble: `Inject secrets as an environment variable:`,
+				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
+				$ hcp vault-secrets run "env"
```
dhuckins, line 60: if we do want to have quotes around the example command, would use single quotes as that doesn't have bash expansion https://stackoverflow.com/a/6697781

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+	result := os.Environ()
+
+	for _, secret := range res.Payload.Secrets {
+		// we need to append results in case of duplicates we want secrets to override
+		// only supporting static secrets
+		if secret.StaticVersion != nil {
+			result = append(result, fmt.Sprintf("%v=%v", strings.ToUpper(secret.Name), secret.StaticVersion.Value))
+
+		}
+	}
```
dhuckins, line 137: if you're using the preview api (2023-11-23 version) `OpenAppSecrets` will return rotating secrets as well
can see how we handled that in `run` here
https://github.com/hashicorp/vlt/blob/b9ce887fec166763cb715bca8fe9be46ee24b3f1/internal/command/run.go#L172-L183

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+		Examples: []cmd.Example{
+			{
+				Preamble: `Inject secrets as an environment variable:`,
+				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
+				$ hcp vault-secrets run "env"
+				`),
+			},
+		},
+		Args: cmd.PositionalArguments{
+			Args: []cmd.PositionalArgument{
```
dhuckins, line 65: does this handle variadic args?
also, do we use the `--` pattern? as in there's the hcp args on the left hand side, but then the command args on the right hand side
eg:
`hcp vault-secrets run -- python3.12 -m http.server`

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func runRun(opts *RunOpts) (err error) {
+	envSecrets, err := getAllSecretsForEnv(opts)
+	if err != nil {
+		return fmt.Errorf("failed to run with secrets in app %q: %w", opts.AppName, err)
+	}
+
+	childProcess := setupChildProcess(opts.Ctx, opts.Command, envSecrets)
+	err = childProcess.Run()
```
dhuckins, line 108: the exit code from the child process isn't captured here

internal/commands/vaultsecrets/run/run.go
dadgar: Yes we do support that and is my suggestion as well. Essentially the arg validation function should be set to [ArbitraryArgs](https://github.com/hashicorp/hcp/blob/main/internal/pkg/cmd/args.go#L23-L25) [here](https://github.com/hashicorp/hcp/blob/main/internal/pkg/cmd/command.go#L146).

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+		PreviewClient: preview_secret_service.New(ctx.HCP, nil),
+		Client:        secret_service.New(ctx.HCP, nil),
+	}
+
+	cmd := &cmd.Command{
+		Name:      "run",
+		ShortHelp: "Run a process with secrets from a Vault Secrets app.",
+		LongHelp: heredoc.New(ctx.IO).Must(`
+		The {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command lets you
+		run the provided command as a child process while injecting
```
dadgar, line 51: The text alignment not being uniform across the lines causes odd help output:

```
DESCRIPTION
  The hcp vault-secrets run command lets you run the provided command as a child
  process while injecting     all of the app's secrets as environment variables,
  with all secret names     converted to upper-case. The stdout/stderr from the
  child     process are forwarded to the top level hcp vault-secrets run command.
```

internal/commands/vaultsecrets/run/run.go
dadgar: Yeah you'll want to set the flag `app` autocomplete not the args

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	childProcess := setupChildProcess(opts.Ctx, opts.Command, envSecrets)
+
+	err = childProcess.Run()
+	if err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			exitCode := exitErr.ExitCode()
+			return fmt.Errorf("failed to run with secrets in app %q; received code %v : %w", opts.AppName, exitCode, err)
```
dadgar, line 124: I just made this PR for you: https://github.com/hashicorp/hcp/pull/103. Would be good to use that so that the exit code gets passed through to the user.

internal/commands/vaultsecrets/run/run_test.go
```diff
@@ -0,0 +1,175 @@
...
+		Name    string
+		Args    []string
+		Profile func(t *testing.T) *profile.Profile
+		Error   string
+		Expect  *RunOpts
+	}{
+		{
+			Name:    "Good",
+			Profile: testProfile,
+			Args:    []string{"env"},
```
dadgar, line 47: Setting the app explicitly via flags would be good

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+		case secret.StaticVersion != nil:
+			result = append(result, fmt.Sprintf("%v=%v", strings.ToUpper(secret.Name), secret.StaticVersion.Value))
+		}
+	}
+
+	return result, nil
+}
+
+func setupChildProcess(ctx context.Context, command []string, envVars []string) *exec.Cmd {
```
dadgar, line 161: Testing this function directly would be good to ensure you are parsing the command and setting the cmd up as expected.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	cmd := &cmd.Command{
+		Name:      "run",
+		ShortHelp: "Run a process with secrets from a Vault Secrets app.",
+		LongHelp: heredoc.New(ctx.IO).Must(`
+		The {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command lets you run 
+		the provided command as a child process while injecting all of the app's secrets
+		as environment variables, with all secret names converted to upper-case. The stdout
+		and stderr from the child process are forwarded to the top level {{ template "mdCodeOrBold" "hcp vault-secrets run" }} command.
```
dadgar, line 56: Would the last sentence be clearer if it was rewritten to say, "STDIN, STDOUT, and STDERR will be passed to the created child process."

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+		`),
+		Examples: []cmd.Example{
+			{
+				Preamble: `Inject secrets as environment variables:`,
```
dadgar, line 60: The `env` example may be nice to still keep. The preamble could maybe allude to it being useful to debug and see the format of the envs.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+			f.Autocomplete = helper.PredictAppName(ctx, cmd, opts.PreviewClient)
+		}
+	}
+
+	return cmd
+}
+
+func runRun(opts *RunOpts) (err error) {
+	if len(opts.Command) == 0 {
+		return fmt.Errorf("failed to run app secrets from name %q - no command provided", opts.AppName)
```
dadgar, line 109:
```suggestion
		return fmt.Errorf("failed to run with secrets in app %q: no command provided", opts.AppName)
```

dadgar (approved):

internal/commands/vaultsecrets/run/run.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	result := os.Environ()
+
+	for _, secret := range res.Payload.Secrets {
+		// we need to append results in case of duplicates we want secrets to override
+		switch {
+		case secret.RotatingVersion != nil:
+			for name, value := range secret.RotatingVersion.Values {
+				result = append(result, fmt.Sprintf("%v_%v=%v", strings.ToUpper(secret.Name), name, value))
```
dhuckins, line 151: for consistency:
```suggestion
				result = append(result, fmt.Sprintf("%v_%v=%v", strings.ToUpper(secret.Name),strings.ToUpper(name), value))
```

dhuckins (approved):
lgtm!

internal/commands/vaultsecrets/run/run.go
mercedesbh: Good catch!

PR #98, opened 2024-05-28
Fix panic displaying invoked command

dadgar:
When the command recieves an error because the authentication token expired and wasn't able to be refreshed, we display a help error informing the user to logout, login, and re-run the command.

A panic was occuring if the previous command had escaped quoted inputs, since the string was being directly added to a template that itself had quotes. This was causing the template string to become invalid, and then rendering the template would error.

This commit causes the command to be properly escaped before it is injected into the template.

### How I've tested this PR:
Added a test that failed before the change, and now works. Also caused the new error message to be displayed for every command being run to verify no more panics.

Does the code logic look correct.

### Checklist:
- [X] Tests added if applicable
- [X] CHANGELOG entry added or label 'pr/no-changelog' added to PR

### DIFF
--- a/.changelog/98.txt
+++ b/.changelog/98.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix rare panic that could occur on authentication error when running a command that had quoted arguments.
+```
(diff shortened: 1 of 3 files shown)


WenInCode (approved):

jasonpilz (approved):
LGTM

PR #118, opened 2024-06-20
Remove module version flags from Waypoint templates & add-ons

paladin-devops:
This PR removes the flags which let users configured the version # of a no-code module for Waypoint templates & add-ons. It also removes the module source flag.

### DIFF
--- a/.changelog/118.txt
+++ b/.changelog/118.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+waypoint: Remove module version flag from templates and add-on definitions commands. Remove module source from update commands.
+```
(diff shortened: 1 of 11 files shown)


catsby (approved):

.changelog/118.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
briancain, line 1: This seems like more of a breaking change than an improvement...

paladin-devops: @briancain I updated the changelog to be a breaking change release note.

PR #119, opened 2024-06-21
Support opening HVS dynamic secrets and enhance formatting

austingebauer:

This PR adds support for opening dynamic secrets in HCP Vault Secrets. Additionally, it enhances the formatting for both dynamic and rotating secrets for format `pretty` and `table`. Prior to this change, the format for the secret values was JSON for both formats. In this PR, the format is `<key>: <value>\n`. Happy to consider other formats or change it back to JSON if you all think that's better.

I'll add/fix up tests once we've got a decision on the formatting.

### How I've tested this PR:

Ran `make go/build` and executed test commands. Also stepped through code with the debugger to fix up the Go templates.

1. Check out this branch
2. Run `make go/build`
3. Run either command
  - `./bin/hcp vault-secrets secrets open --format=pretty <your-dynamic-secret-name>`
  - `./bin/hcp vault-secrets secrets open --format=table <your-dynamic-secret-name>`

## Example output:

```shell
$ ./bin/hcp vault-secrets secrets open --format=table gcp_service_account_reader
Secret Name                  Latest Version   Created At                 Type      Values                                                  
gcp_service_account_reader   0                2024-06-20T23:43:45.264Z   dynamic   access_token: Redacted
                                                                                   token_type: Bearer

$ ./bin/hcp vault-secrets secrets open --format=pretty gcp_service_account_reader
Secret Name:    gcp_service_account_reader
Latest Version: 0
Created At:     2024-06-20T23:43:45.264Z
Type:           dynamic
Values:         access_token: Redacted
                token_type: Bearer
```

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/secrets/open.go
+++ b/internal/commands/vaultsecrets/secrets/open.go
@@ -37,7 +37,7 @@ func NewCmdOpen(ctx *cmd.Context, runF func(*OpenOpts) error) *cmd.Command {
 		Name:      "open",
 		ShortHelp: "Open a static, rotating, or dynamic secret.",
 		LongHelp: heredoc.New(ctx.IO).Must(`
-		The {{ template "mdCodeOrBold" "hcp vault-secrets secrets open" }} command reads the plaintext value of a static secret from the Vault Secrets application.
+		The {{ template "mdCodeOrBold" "hcp vault-secrets secrets open" }} command reads the plaintext value of a static, rotating, or dynamic secret from the Vault Secrets application.
 		`),
 		Examples: []cmd.Example{
 			{
--- a/internal/commands/vaultsecrets/secrets/open.go
+++ b/internal/commands/vaultsecrets/secrets/open.go
@@ -35,7 +35,7 @@ func NewCmdOpen(ctx *cmd.Context, runF func(*OpenOpts) error) *cmd.Command {
 
 	cmd := &cmd.Command{
 		Name:      "open",
-		ShortHelp: "Open a static secret.",
+		ShortHelp: "Open a static, rotating, or dynamic secret.",
 		LongHelp: heredoc.New(ctx.IO).Must(`
 		The {{ template "mdCodeOrBold" "hcp vault-secrets secrets open" }} command reads the plaintext value of a static secret from the Vault Secrets application.
 		`),
--- a/internal/commands/vaultsecrets/secrets/open.go
+++ b/internal/commands/vaultsecrets/secrets/open.go
@@ -35,7 +35,7 @@ func NewCmdOpen(ctx *cmd.Context, runF func(*OpenOpts) error) *cmd.Command {
 
 	cmd := &cmd.Command{
 		Name:      "open",
-		ShortHelp: "Open a static, rotating, or dynamic secret.",
+		ShortHelp: "Open a secret.",
 		LongHelp: heredoc.New(ctx.IO).Must(`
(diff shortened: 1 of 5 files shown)


maxcoulombe (approved):
Nice, I like the enhanced format for multi-value secrets.

Test, changelog and ship it!

maxcoulombe:
A bit out of scope of this PR but the helper text for the read and open commands mention static secrets while they apply for all 3 types:
```
COMMANDS
  create:  Create a new static secret.
  delete:  Delete a static secret.
  list:    List an application's secrets.
  open:    Open a static secret.
  read:    Read a static secret's metadata.
```
We could fix it here or in a follow-up PR.

averche:
You may also want to update run.go here:

https://github.com/hashicorp/hcp/blob/4b390db85940e3df09cfe15da7838f5211d256d9/internal/commands/vaultsecrets/run/run.go#L155-L161

Also curious if the `-format=json` is going display the unredacted secret?

austingebauer:
@averche - Good call, will update `run.go`. 

> Curious if the -format=json is going to print the unredacted secret?

Yep, it prints the unredacted secret in valid JSON as you'd expect.

internal/commands/vaultsecrets/secrets/open.go
```diff
@@ -36,7 +35,7 @@ func NewCmdOpen(ctx *cmd.Context, runF func(*OpenOpts) error) *cmd.Command {
 
 	cmd := &cmd.Command{
 		Name:      "open",
-		ShortHelp: "Open a static secret.",
+		ShortHelp: "Open a static, rotating, or dynamic secret.",
 		LongHelp: heredoc.New(ctx.IO).Must(`
 		The {{ template "mdCodeOrBold" "hcp vault-secrets secrets open" }} command reads the plaintext value of a static secret from the Vault Secrets application.
```
dhuckins, line 40: probably need to update the long help too

internal/commands/vaultsecrets/secrets/open.go
```diff
@@ -120,42 +119,55 @@ func openRun(opts *OpenOpts) (err error) {
...
 	case resp.Payload.Secret.RotatingVersion != nil:
 		secretValue, err = formatSecretMap(resp.Payload.Secret.RotatingVersion.Values)
 		if err != nil {
 			secretValue = "<< COULD NOT ENCODE TO JSON >>"
 		}
+
+		field = format.Field{
+			Name:        "Values",
+			ValueFormat: `{{ range $key, $value := .RotatingVersion.Values }}{{printf "%s: %s\n" $key $value}}{{ end }}`,
+		}
```
dhuckins, line 137: if `formatSecretMap` encodes the map to JSON, why do we also need the format field here?

internal/commands/vaultsecrets/secrets/open.go
austingebauer: Whoops! Done in https://github.com/hashicorp/hcp/pull/119/commits/40bb61c0989585ea52825e43f634726d71fb1e43.

internal/commands/vaultsecrets/secrets/open.go
austingebauer: The output of `formatSecretMap` is assigned to `secretValue` which is only used when the secret values are written to a file. `secretValue` is not used otherwise.

The format field here defines how we want to template the secret values for each type when its displayed in the terminal.

dadgar (approved):
Nice use of the field template

PR #127, opened 2024-07-03
[HVS] Warn on run collision

VinnyHC:
When running `hcp vault-secrets run` return an error if the fomated secret names collide in the prepared environment.

https://hashicorp.atlassian.net/browse/VAULT-28641

Screenshot 2024-07-08 at 2 28 49 PM

### How I've tested this PR:
New tests have been added.
Tested against issue in HCP.

Set up conflicting secret names in an app and issue a `hcp vault-secrets run` command to see error message.

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/run/run.go
+++ b/internal/commands/vaultsecrets/run/run.go
@@ -178,7 +178,12 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
 		if !collided {
 			delete(collisions, fmtName)
 		} else {
-			opts.Logger.Warn(fmt.Sprintf("environment variable \"%s\" was assigned more than once", fmtName))
+			_, err = fmt.Fprintf(opts.IO.Err(), "%s Environment variable \"%s\" was assigned more than once\n",
+				opts.IO.ColorScheme().WarningLabel(), fmtName)
+			if err != nil {
+				return nil, err
+			}
+
 		}
 	}
 
--- a/internal/commands/vaultsecrets/run/run.go
+++ b/internal/commands/vaultsecrets/run/run.go
@@ -179,16 +179,12 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
 	for fmtName, uses := range collisions {
 		if len(uses) > 1 {
 			hasCollisions = true
-			var offenders string
+			offenders := make([]string, len(uses))
 			for i, use := range uses {
-				offenders += fmt.Sprintf("\"%s\" [%s]", use.Name, use.Type)
-				if len(uses)-1 != i {
-					offenders += ", "
-				}
+				offenders[i] = fmt.Sprintf("\"%s\" [%s]", use.Name, use.Type)
 			}
-
 			_, err = fmt.Fprintf(opts.IO.Err(), "%s %s map to the same environment variable \"%s\"\n",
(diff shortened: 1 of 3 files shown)


internal/commands/vaultsecrets/run/run.go
```diff
@@ -148,26 +149,55 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
...
+				return nil, err
+			}
+
 		}
 	}
 
 	return result, nil
 }
 
+func processCollisions(collisions map[string]bool, fmtName string) {
```
ncabatoff, line 193: You could eliminate the conditional (and thereby the need for the func processCollisions) if you made collisions a map from string to int.  Then instead of doing `processCollisions(collisions, fmtName)`, you could just do `collisions[fmtName]++`.

internal/commands/vaultsecrets/run/run.go
```diff
@@ -148,26 +149,55 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
...
+			result = append(result, fmt.Sprintf("%v=%v", fmtName, secret.StaticVersion.Value))
+		}
+	}
+
+	// check collisions by dropping unique records from the map
+	for fmtName, collided := range collisions {
+		if !collided {
+			delete(collisions, fmtName)
+		} else {
+			_, err = fmt.Fprintf(opts.IO.Err(), "%s Environment variable \"%s\" was assigned more than once\n",
```
ncabatoff, line 181: Why only a warning as opposed to an error?

ncabatoff (approved):

internal/commands/vaultsecrets/run/run.go
averche: Agreed, error here may be more appropriate. 

Would it be also useful to display the secrets that contributed to the collision (can keep track of them in the same map)? Example:

```go
ERROR: secrets "mongodb" [rotating], "mongodb_username" [static] map to the same environment variable "MONGODB_USERNAME"
```

averche (approved):

internal/commands/vaultsecrets/run/run.go
VinnyHC: Used this as inspiration in 05baedf

internal/commands/vaultsecrets/run/run.go
```diff
@@ -149,53 +150,56 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
...
+	hasCollisions := false
+	for fmtName, uses := range collisions {
+		if len(uses) > 1 {
+			hasCollisions = true
+			var offenders string
+			for i, use := range uses {
+				offenders += fmt.Sprintf("\"%s\" [%s]", use.Name, use.Type)
+				if len(uses)-1 != i {
+					offenders += ", "
+				}
```
averche, line 187: nit: this can be a tiny bit cleaner if `offenders` is a `[]string` and we use `strings.Join(offenders, ", ")`

averche (approved):

internal/commands/vaultsecrets/run/run.go
VinnyHC: 🧠  3f10ceb

austingebauer (approved):
Nice!

PR #130, opened 2024-07-19
Waypoint: Support for create/update Templates with Input Variables

catsby:

- Adds a new field `--variable-options-file` which accepts a properly formatted HCL file of Template Variable options. Tests are also added/modified to verify the field. 

### How I've tested this PR:

Testing this requires a HCP Waypoint Template with variable options available, which itself requires a Terraform no-code Module to back it, so it's not the most straightforward thing to setup.

Example variable file:

```hcl
// vars.hcl
variable_option "string_variable" {
  type = "string"
  options = [
    "a string value",
  ]
  user_editable = false
}

variable_option "misc_variable" {
  type = "string"
  options = [
    "another string value",
  ]
  user_editable = false
}
```

Example command usage:

```shell
➜ hcp waypoint templates update -n=cts-null-template-from-cli --variable-options-file=vars.hcl
```

### Checklist:


paladin-devops (approved):

internal/commands/profile/init.go
```diff
@@ -302,11 +302,11 @@ func (i *InitOpts) gatherProjectID(detectedProject string) (string, error) {
 	if err != nil {
 		var listErr *project_service.ProjectServiceListDefault
 		if errors.As(err, &listErr) && listErr.IsCode(http.StatusForbidden) && detectedProject != "" {
-			fmt.Fprintf(i.IO.Err(), heredoc.New(i.IO).Mustf(`
-			{{ Color "yellow" "Principal does not have permission to list projects." }}
+			fmt.Fprintf(i.IO.Err(), "%s\n", heredoc.New(i.IO).Mustf(`
```
dadgar, line 305: Should this become `fmt.Fprintln(i.IO.Err(), heretic.New(i.IO).Mustf(`....`))`?

PR #137, opened 2024-07-26
VAULT-29259 -  List HVS integrations

mercedesbh:
This PR adds the `list` command for hvs apps for ticket https://hashicorp.atlassian.net/browse/VAULT-29259

### How I've tested this PR:

1. Checkout this branch
2. Run `make go/build`
3. Read an integration `./bin/hcp vault-secrets integrations list --type {YOUR_INTEGRATION_TYPE}` or `./bin/hcp vault-secrets integrations list`

Screenshot 2024-07-30 at 4 32 53 PM

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/integrations/displayer.go
+++ b/internal/commands/vaultsecrets/integrations/displayer.go
@@ -47,7 +47,7 @@ func (t *twilioDisplayer) FieldTemplates() []format.Field {
 	fields := []format.Field{
 		{
 			Name:        "Integration Name",
-			ValueFormat: "{{ .IntegrationName }}",
+			ValueFormat: "{{ .Name }}",
 		},
 	}
 
@@ -103,7 +103,7 @@ func (m *mongodbDisplayer) FieldTemplates() []format.Field {
 	fields := []format.Field{
 		{
 			Name:        "Integration Name",
-			ValueFormat: "{{ .IntegrationName }}",
+			ValueFormat: "{{ .Name }}",
 		},
 	}
 
@@ -158,7 +158,7 @@ func (g *genericDisplayer) FieldTemplates() []format.Field {
 	return []format.Field{
 		{
 			Name:        "Integration Name",
-			ValueFormat: "{{ .IntegrationName }}",
+			ValueFormat: "{{ .Name }}",
 		},
 	}
 }
--- a/internal/commands/vaultsecrets/integrations/displayer.go
+++ b/internal/commands/vaultsecrets/integrations/displayer.go
@@ -55,7 +55,11 @@ func (t *twilioDisplayer) FieldTemplates() []format.Field {
 		return append(fields, []format.Field{
 			{
 				Name:        "Account SID",
-				ValueFormat: "{{ .TwilioAccountSid }}",
+				ValueFormat: "{{ .StaticCredentialDetails.AccountSid }}",
+			},
+			{
+				Name:        "API Key SID",
+				ValueFormat: "{{ .StaticCredentialDetails.APIKeySid }}",
 			},
 		}...)
 	} else {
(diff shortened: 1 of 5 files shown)


internal/commands/vaultsecrets/integrations/list.go
```diff
@@ -0,0 +1,163 @@
...
+func listRun(opts *ListOpts) error {
+	if opts.Type == "" {
+		var integrations []*models.Secrets20231128Integration
+		params := &preview_secret_service.ListIntegrationsParams{
+			Context:        opts.Ctx,
+			ProjectID:      opts.Profile.ProjectID,
+			OrganizationID: opts.Profile.OrganizationID,
+		}
+		for {
+			resp, err := opts.PreviewClient.ListIntegrations(params, nil)
```
mercedesbh, line 91: Not sure why I'm getting a 404 when hitting `ListIntegrations`. Even when trying directly by hitting the API. Lmk if anyone has ideas!
Screenshot 2024-07-30 at 4 35 54 PM
Screenshot 2024-07-30 at 4 36 08 PM

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -44,15 +44,22 @@ func (t *twilioDisplayer) previewTwilioIntegrationsPayload() any {
...
 			ValueFormat: "{{ .IntegrationName }}",
 		},
-		{
-			Name:        "Account SID",
-			ValueFormat: "{{ .TwilioAccountSid }}",
-		},
+	}
+
+	if t.single {
+		return append(fields, []format.Field{
```
pmmukh, line 55: curious, why does this go in only if a single integration is being shown? Shouldn't `TwilioAccountSid` be in all the response entries?

internal/commands/vaultsecrets/integrations/list.go
```diff
@@ -0,0 +1,163 @@
...
+			{
+				Preamble: `List twilio integrations:`,
+				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
+				$ hcp vault-secrets integrations list --type "twilio"
+				`),
+			},
+			{
+				Preamble: `List all generic integrations:`,
+				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
+				$ hcp vault-secrets integrations list"
```
pmmukh, line 58:
```suggestion
				$ hcp vault-secrets integrations list
```

internal/commands/vaultsecrets/integrations/list.go
pmmukh: Nothing helpful to add :( it seems like you're doing the right thing, and the endpoint _should_ be querying from the third_party_integrations table

internal/commands/vaultsecrets/integrations/list_test.go
```diff
@@ -0,0 +1,164 @@
...
+		})
+	}
+}
+
+func TestListRun(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		Name   string
+		ErrMsg string
```
pmmukh, line 88: should a test case exercising `ErrMsg` be added here? I think i'd vote for either that or removing the `ErrMsg` field

internal/commands/vaultsecrets/integrations/displayer.go
averche: We currently don't populate the credential details on LIST requests (we may add them later but there were some performance concerns with queries to tokenization)

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -44,15 +44,22 @@ func (t *twilioDisplayer) previewTwilioIntegrationsPayload() any {
 }
 
 func (t *twilioDisplayer) FieldTemplates() []format.Field {
-	return []format.Field{
+	fields := []format.Field{
 		{
 			Name:        "Integration Name",
 			ValueFormat: "{{ .IntegrationName }}",
```
averche, line 50: This field will be deprecated in favor of `Name`

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -44,15 +44,22 @@ func (t *twilioDisplayer) previewTwilioIntegrationsPayload() any {
...
-			Name:        "Account SID",
-			ValueFormat: "{{ .TwilioAccountSid }}",
-		},
+	}
+
+	if t.single {
+		return append(fields, []format.Field{
+			{
+				Name:        "Account SID",
+				ValueFormat: "{{ .TwilioAccountSid }}",
```
averche, line 58: This field will be deprecated, we have a nested StaticCredentialDetails which will contain provider-specific fields

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -44,15 +44,22 @@ func (t *twilioDisplayer) previewTwilioIntegrationsPayload() any {
...
-		{
-			Name:        "Account SID",
-			ValueFormat: "{{ .TwilioAccountSid }}",
-		},
+	}
+
+	if t.single {
+		return append(fields, []format.Field{
+			{
+				Name:        "Account SID",
```
averche, line 57: Do we also want to display the Twilio API Key SID?

internal/commands/vaultsecrets/integrations/list.go
```diff
@@ -0,0 +1,163 @@
...
+			if resp.Payload.Pagination == nil || resp.Payload.Pagination.NextPageToken == "" {
+				break
+			}
+
+			next := resp.Payload.Pagination.NextPageToken
+			params.PaginationNextPageToken = &next
+		}
+		return opts.Output.Display(newMongoDBDisplayer(false, integrations...))
+
+	default:
```
averche, line 160: We may also want to support AWS & GCP here, they follow the same pattern

internal/commands/vaultsecrets/integrations/list.go
averche, line 91: ListIntegrations also supports filtering by capabilities (e.g. ROTATION/DYNAMIC), not sure if we want to support it now just FYI.

averche: Just tested this locally and it seems to worked for me, happy to help debug this if its still an issue

mercedesbh: Not sure what the issue is here, I'll take some time to investigate deeper when I get a bit more time in the project. But for now I was able to confirm the general list works against my personal environment, and Anton was able to confirm it worked against prod for him (checked out this branch).

mercedesbh: Sounds good! Will add a ticket to add the filtering by capabilities in the project's third milestone.

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -93,14 +104,181 @@ func (m *mongodbDisplayer) previewMongoDBIntegrationsPayload() any {
...
+			Name:        "Integration Name",
+			ValueFormat: "{{ .Name }}",
+		},
+	}
+
+	if a.single {
+		return append(fields, []format.Field{
+			{
+				Name:        "Audience",
+				ValueFormat: "{{ .FederatedWorkloadIdentity.Audience }}",
```
averche, line 174: I believe we also return Role ARN in the response, not sure if you want to also display it.

internal/commands/vaultsecrets/integrations/displayer.go
```diff
@@ -93,14 +104,181 @@ func (m *mongodbDisplayer) previewMongoDBIntegrationsPayload() any {
...
+			Name:        "Integration Name",
+			ValueFormat: "{{ .Name }}",
+		},
+	}
+
+	if g.single {
+		return append(fields, []format.Field{
+			{
+				Name:        "Audience",
+				ValueFormat: "{{ .FederatedWorkloadIdentity.Audience }}",
```
averche, line 230: AWS and GCP have a couple authentication methods each now:
- FederatedWorkloadIdentity
- Static Credentials (Access Key for AWS & Service Account Key for GCP)

The two authentication methods are mutually exclusive, meaning that if `.ServiceAccountKey` is populated, `FederatedWorkloadIdentity` will not be and wise versa.

I'm not sure how to handle this in the display layer but something to keep in mind.

averche (approved):
A couple small comments, o/w LGTM!

internal/commands/vaultsecrets/integrations/displayer.go
mercedesbh: Ah interesting, thanks for this info. I'll find a way to address this in the displayer.

PR #140, opened 2024-08-02
VAULT-29012 - Create HVS integrations

mercedesbh:
This PR adds the create command for hvs apps for ticket https://hashicorp.atlassian.net/browse/VAULT-29012

### How I've tested this PR:

1. Checkout this branch
2. Run `make go/build`
3. Create an integration `./bin/hcp vault-secrets integrations create {INTEGRATION_NAME} --config-file "/path/to/file/config.yaml"`

Config file `integration-config.yaml`
Screenshot 2024-08-02 at 3 59 45 PM
Screenshot 2024-08-02 at 3 59 17 PM

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/integrations/create.go
+++ b/internal/commands/vaultsecrets/integrations/create.go
@@ -6,9 +6,14 @@ package integrations
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/models"
+	"os"
+	"slices"
+
+	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v3"
 
 	preview_secret_service "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/client/secret_service"
+	preview_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/stable/2023-06-13/client/secret_service"
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
 	"github.com/hashicorp/hcp/internal/pkg/flagvalue"
@@ -16,8 +21,6 @@ import (
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
 	"github.com/hashicorp/hcp/internal/pkg/iostreams"
 	"github.com/hashicorp/hcp/internal/pkg/profile"
-	"gopkg.in/yaml.v3"
-	"os"
 )
 
 type CreateOpts struct {
@@ -72,6 +75,7 @@ func NewCmdCreate(ctx *cmd.Context, runF func(*CreateOpts) error) *cmd.Command {
 					DisplayValue: "CONFIG_FILE",
 					Description:  "File path to read integration config data.",
 					Value:        flagvalue.Simple("", &opts.ConfigFilePath),
+					Required:     true,
 				},
 			},
 		},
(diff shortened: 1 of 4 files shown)


internal/commands/vaultsecrets/integrations/create.go
```diff
@@ -0,0 +1,248 @@
...
+	case Twilio:
+		missingField := validateDetails(i.Details, TwilioKeys)
+
+		if missingField != "" {
+			return fmt.Errorf("missing required field in the config file: %s", missingField)
+		}
+
+		body := &preview_models.SecretServiceCreateTwilioIntegrationBody{
+			Name:               opts.IntegrationName,
+			TwilioAccountSid:   i.Details["account_sid"],
```
pmmukh, line 131: thoughts on doing `TwilioKeys[0]` etc. for these fields, instead of `account_sid`? and similarly for the others

internal/commands/vaultsecrets/integrations/create.go
```diff
@@ -0,0 +1,248 @@
...
+			ProjectID:      opts.Profile.ProjectID,
+			OrganizationID: opts.Profile.OrganizationID,
+			Body:           body,
+		}, nil)
+
+		if err != nil {
+			return fmt.Errorf("failed to create Twilio integration: %w", err)
+		}
+
+		fmt.Fprintln(opts.IO.Err())
```
pmmukh, line 147: thoughts on moving  these 2 fmt.Fprint lines out of the switch cases and before the return? since it can also be done with opts.IntegrationName, and should reduce lines a bit

internal/commands/vaultsecrets/integrations/create.go
```diff
@@ -0,0 +1,248 @@
...
+		fmt.Fprintf(opts.IO.Err(), "%s Successfully created integration with name %q\n", opts.IO.ColorScheme().SuccessIcon(), resp.Payload.Integration.Name)
+	}
+
+	return nil
+}
+
+func validateDetails(details map[string]string, requiredKeys []string) string {
```
pmmukh, line 239: thoughts on returning an array here with all missing keys?

internal/commands/vaultsecrets/integrations/create_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+			r.Zero(code, io.Error.String())
+			r.NotNil(gotOpts)
+			r.Equal(c.Expect.IntegrationName, gotOpts.IntegrationName)
+			r.Equal(c.Expect.ConfigFilePath, gotOpts.ConfigFilePath)
+		})
+	}
+}
+
+func TestCreateRun(t *testing.T) {
```
pmmukh, line 101: this looks like a twilio specific test, so maybe a rename to `TestCreateTwilioRun` might make sense?

pmmukh (commented):
looking great!!

internal/commands/vaultsecrets/integrations/create.go
mercedesbh: Great suggestion!
Screenshot 2024-08-13 at 5 11 20 PM

internal/commands/vaultsecrets/integrations/create.go
mercedesbh: Hmmm, that's a good point. I'll include 2 other providers in the test cases.

internal/commands/vaultsecrets/integrations/create_test.go
mercedesbh: Hmmm, that's a good point. I'll include 2 other providers in the test cases.

pmmukh: thanks!!

PR #141, opened 2024-08-07
VAULT-29260 - Create rotating secrets

mercedesbh:
This PR edits the secrets create command for hvs rotating secrets for ticket https://hashicorp.atlassian.net/browse/VAULT-29260

### How I've tested this PR:

1. Checkout this branch
2. Run `make go/build`
3. Create a rotating secret `./bin/hcp vault-secrets secrets create {SECRET_NAME} --secret-type rotating --data-file "/path/to/file/config.yaml"`

Example config file:
```
type: "twilio"
rotation_integration_name: "Twil-Int-4"
details:
  rotation_policy_name: "60"
```

Screenshot 2024-08-16 at 3 11 26 PM

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/secrets/create.go
+++ b/internal/commands/vaultsecrets/secrets/create.go
@@ -7,15 +7,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
-	"github.com/mitchellh/mapstructure"
-	"gopkg.in/yaml.v3"
 	"io"
 	"os"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/posener/complete"
+	"gopkg.in/yaml.v3"
 
 	preview_secret_service "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/client/secret_service"
 	preview_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/stable/2023-06-13/client/secret_service"
+	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
 	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/secrets/appname"
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
 	"github.com/hashicorp/hcp/internal/pkg/flagvalue"
@@ -23,7 +25,6 @@ import (
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
 	"github.com/hashicorp/hcp/internal/pkg/iostreams"
 	"github.com/hashicorp/hcp/internal/pkg/profile"
-	"github.com/posener/complete"
 )
 
 type SecretType string
--- a/internal/commands/vaultsecrets/secrets/create.go
+++ b/internal/commands/vaultsecrets/secrets/create.go
@@ -202,10 +202,10 @@ func createRun(opts *CreateOpts) error {
 
 		switch sc.Type {
 		case integrations.Twilio:
-			missingField := validateDetails(sc.Details, TwilioKeys)
+			missingFields := validateDetails(sc.Details, TwilioKeys)
 
-			if missingField != "" {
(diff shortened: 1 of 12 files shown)


internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -98,35 +120,177 @@ type CreateOpts struct {
...
+}
+
+type MongoDBScope struct {
+	Name string `mapstructure:"type"`
+	Type string `mapstructure:"name"`
+}
+
+var (
+	TwilioKeys = []string{"rotation_policy_name"}
+	MongoKeys  = []string{"rotation_policy_name", "mongodb_group_id"}
```
averche, line 148: rotation_policy_name should be required for all rotating secrets (just like `rotation_integration_name`

For MongoDB, roles is also required

internal/commands/vaultsecrets/secrets/create_test.go
```diff
@@ -45,12 +50,21 @@ func TestNewCmdCreate(t *testing.T) {
...
 			Profile: testProfile,
 			Args:    []string{},
-			Error:   "ERROR: accepts 1 arg(s), received 0",
+			Error:   "ERROR: missing required flag: --data-file=DATA_FILE_PATH",
 		},
 		{
 			Name:    "Good: Secret name arg specified",
 			Profile: testProfile,
-			Args:    []string{"test"},
+			Args:    []string{"test", "--data-file=DATA_FILE_PATH"},
```
averche, line 58: Could you please show an example of what a data file for Twilio/MongoDB would look like (probably worth adding it to the PR description and maybe the documentation in another PR).

averche, line 58: Curious if this also supports [heredoc](https://en.wikipedia.org/wiki/Here_document) input

dhuckins: +1

dhuckins (commented):
lgtm! think @averche covered everything

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: `rotation_integration_name` is added here for these two separately because the `SecretConfig` is shared w/ dynamic secrets as well, which [don't have `rotation_integration_name`](https://github.com/hashicorp/hcp/pull/143/files#diff-7f92b17a95650abeca3ec432e8bcf6859f19709d1bdb6b3687fb5d704940a4ecR146-R150). I could see an argument for making a separating the two though.

averche: Gotcha, it's not a big deal then. If any required parameter is missing, the request will be rejected by server side validation

internal/commands/vaultsecrets/secrets/create_test.go
mercedesbh: I can def look at this as part of Milestone 3 (toward the project's end). That would likely be useful for many

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -98,35 +120,173 @@ type CreateOpts struct {
...
+}
+
+type MongoDBScope struct {
+	Name string `mapstructure:"type"`
+	Type string `mapstructure:"name"`
+}
+
+var (
+	// There are no Twilio-specific keys
+	MongoKeys = []string{"mongodb_group_id", "mongodb_roles"}
```
averche, line 149: Nit: Maybe call these MongoDBAtlasRequireKeys? (since there are other keys like `mongodb_scopes`)

averche (approved):
LGTM!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -67,11 +82,18 @@ func NewCmdCreate(ctx *cmd.Context, runF func(*CreateOpts) error) *cmd.Command {
 					DisplayValue: "DATA_FILE_PATH",
 					Description:  "File path to read secret data from. Set this to '-' to read the secret data from stdin.",
 					Value:        flagvalue.Simple("", &opts.SecretFilePath),
+					Required:     true,
```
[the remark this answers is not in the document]
mercedesbh: Yep, it is. Not sure why it wasn't already marked as required, but even for static credentials the value can't be passed in as plaintext (unlike vlt)

PR #143, opened 2024-08-14
VAULT-29787 - Create dynamic secrets

mercedesbh:
This PR adds the create command for hvs apps for ticket https://hashicorp.atlassian.net/browse/VAULT-29787

### How I've tested this PR:
Ran 1make go/build1 to manually test local changes, and added unit tests.

1. Checkout this branch
2. Run `make go/build`
3. Create a rotating secret `./bin/hcp vault-secrets secrets create {SECRET_NAME} --secret-type dynamic --data-file "/path/to/file/config.yaml"`

Screenshot 2024-08-29 at 3 20 10 PM

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/secrets/create.go
+++ b/internal/commands/vaultsecrets/secrets/create.go
@@ -7,19 +7,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
+	"github.com/mitchellh/mapstructure"
+	"gopkg.in/yaml.v3"
 	"io"
 	"os"
 	"slices"
 
-	"github.com/mitchellh/mapstructure"
-	"github.com/posener/complete"
-	"golang.org/x/exp/maps"
-	"gopkg.in/yaml.v3"
-
 	preview_secret_service "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/client/secret_service"
 	preview_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/stable/2023-06-13/client/secret_service"
-	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
 	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/secrets/appname"
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
 	"github.com/hashicorp/hcp/internal/pkg/flagvalue"
@@ -27,6 +24,8 @@ import (
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
 	"github.com/hashicorp/hcp/internal/pkg/iostreams"
 	"github.com/hashicorp/hcp/internal/pkg/profile"
+	"github.com/posener/complete"
+	"golang.org/x/exp/maps"
 )
 
 type SecretType string
@@ -198,7 +197,6 @@ func createRun(opts *CreateOpts) error {
 
 		switch sc.Type {
 		case integrations.Twilio:
-
 			req := preview_secret_service.NewCreateTwilioRotatingSecretParamsWithContext(opts.Ctx)
 			req.OrganizationID = opts.Profile.OrganizationID
 			req.ProjectID = opts.Profile.ProjectID
--- a/internal/commands/vaultsecrets/secrets/create.go
+++ b/internal/commands/vaultsecrets/secrets/create.go
@@ -7,16 +7,19 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
-	"github.com/mitchellh/mapstructure"
-	"gopkg.in/yaml.v3"
 	"io"
 	"os"
 	"slices"
 
+	"github.com/mitchellh/mapstructure"
+	"github.com/posener/complete"
+	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v3"
+
 	preview_secret_service "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/client/secret_service"
 	preview_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/preview/2023-11-28/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-secrets/stable/2023-06-13/client/secret_service"
+	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/integrations"
 	"github.com/hashicorp/hcp/internal/commands/vaultsecrets/secrets/appname"
 	"github.com/hashicorp/hcp/internal/pkg/cmd"
 	"github.com/hashicorp/hcp/internal/pkg/flagvalue"
@@ -24,8 +27,6 @@ import (
 	"github.com/hashicorp/hcp/internal/pkg/heredoc"
 	"github.com/hashicorp/hcp/internal/pkg/iostreams"
 	"github.com/hashicorp/hcp/internal/pkg/profile"
-	"github.com/posener/complete"
-	"golang.org/x/exp/maps"
 )
 
 type SecretType string
@@ -143,8 +144,12 @@ type MongoDBScope struct {
 	Type string `mapstructure:"name"`
 }
 
-// There are no Twilio-specific keys
-var MongoDBAtlasRequiredKeys = []string{"mongodb_group_id", "mongodb_roles"}
+var (
+	// There are no Twilio-specific keys
+	MongoDBAtlasRequiredKeys = []string{"mongodb_group_id", "mongodb_roles"}
+	AwsKeys    = []string{"default_ttl", "role_arn"}
+	GcpKeys    = []string{"default_ttl", "service_account_email"}
+)
 
 var rotationPolicies = map[string]string{
 	"30": "built-in:30-days-2-active",
@@ -178,16 +183,15 @@ func createRun(opts *CreateOpts) error {
 			return err
 		}
 	case Rotating:
-		f, err := os.ReadFile(opts.SecretFilePath)
+		sc, err := readConfigFile(opts)
 		if err != nil {
-			return fmt.Errorf("failed to open config file: %w", err)
+			return fmt.Errorf("failed to process config file: %w", err)
 		}
 
 		var sc RotatingSecretConfig
 		err = yaml.Unmarshal(f, &sc)
 		if err != nil {
 			return fmt.Errorf("failed to unmarshal config file: %w", err)
-		}
 
 		missingFields := validateRotatingSecretConfig(sc)
 
@@ -197,6 +201,12 @@ func createRun(opts *CreateOpts) error {
 
 		switch sc.Type {
 		case integrations.Twilio:
+			missingDetails := validateDetails(sc.Details, TwilioKeys)
+
+			if len(missingDetails) > 0 {
+				return fmt.Errorf("missing required field(s) in the config file details: %s", missingDetails)
+			}
+
 			req := preview_secret_service.NewCreateTwilioRotatingSecretParamsWithContext(opts.Ctx)
 			req.OrganizationID = opts.Profile.OrganizationID
 			req.ProjectID = opts.Profile.ProjectID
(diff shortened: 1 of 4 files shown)


mercedesbh:
Some messy rebasing removed some changes...fixing now.

austingebauer:
@mercedesbh - Letting you know I see this and will review it tomorrow 👀 Ty!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -288,6 +281,70 @@ func createRun(opts *CreateOpts) error {
 				return err
 			}
 		}
+	case Dynamic:
```
austingebauer, line 284: There are constants for secret types in [secret_types.go](https://github.com/hashicorp/hcp/blob/main/internal/commands/vaultsecrets/secrets/secret_types.go). Would be good to use one set of them instead of having two?

austingebauer (commented):
Would we be updating the help text in this PR or another one? Currently it mentioned static secrets but not dynamic/rotating:
```
$ ./bin/hcp vs secrets create --help      
USAGE
  hcp vault-secrets secrets create NAME --data-file=DATA_FILE_PATH [Optional
    Flags]

DESCRIPTION
  The hcp vault-secrets secrets create command creates a new static secret under
  a Vault Secrets application.
```

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -341,6 +398,35 @@ func readPlainTextSecret(opts *CreateOpts) error {
...
+
+func validateFields(sc SecretConfig) []string {
+	var missingKeys []string
+
+	if sc.Type == "" || sc.IntegrationName == "" {
+		missingKeys = append(missingKeys, "type")
+	}
+
+	if sc.IntegrationName == "" {
+		missingKeys = append(missingKeys, "rotation_integration_name")
```
austingebauer, line 424: Same here, I believe the validation would fail for dynamic secrets given it doesn't have a `rotation_integration_name` field?

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -288,6 +281,70 @@ func createRun(opts *CreateOpts) error {
...
+			req := preview_secret_service.NewCreateAwsDynamicSecretParamsWithContext(opts.Ctx)
+			req.OrganizationID = opts.Profile.OrganizationID
+			req.ProjectID = opts.Profile.ProjectID
+			req.AppName = opts.AppName
+			req.Body = &preview_models.SecretServiceCreateAwsDynamicSecretBody{
+				IntegrationName: sc.IntegrationName,
+				DefaultTTL:      sc.Details[AwsKeys[0]].(string),
+				AssumeRole: &preview_models.Secrets20231128AssumeRoleRequest{
+					RoleArn: sc.Details[AwsKeys[1]].(string),
+				},
```
austingebauer, line 313: I wonder if we'd be able to unmarshal the serialized data from the file directly into the API structures like `preview_models.Secrets20231128AssumeRoleRequest`? Relying on the order of values in `AwsKeys` seems like it could become error prone.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -146,6 +146,8 @@ type MongoDBScope struct {
 var (
 	TwilioKeys = []string{"rotation_policy_name"}
 	MongoKeys  = []string{"rotation_policy_name", "mongodb_group_id"}
+	AwsKeys    = []string{"default_ttl", "role_arn"}
+	GcpKeys    = []string{"default_ttl", "service_account_email"}
```
austingebauer, line 150: Happen to have an example of the YAML structure for say AWS? Trying to think if it makes sense to group related fields here similar to how we do in the API. 

For example, the AWS dynamic secret create API has a request body like:
```json
{
  "name": "s3_reader",
  "integration_name": "aws-integration",
  "default_ttl": "3600s",
  "assume_role": {
    "role_arn": "arn:aws:iam::644381835574:role/hvs-aws-s3-lister-prde"
  }
}
```

The `assume_role` field groups related fields. In the future, we might have an `iam_user` grouping for a different credential type.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -288,6 +281,70 @@ func createRun(opts *CreateOpts) error {
...
+			return fmt.Errorf("failed to process config file: %w", err)
+		}
+
+		missingFields := validateFields(sc)
+		if len(missingFields) > 0 {
+			return fmt.Errorf("missing required field(s) in the config file: %s", missingFields)
+		}
+
+		switch sc.Type {
+		case integrations.AWS:
```
austingebauer, line 296: The switch cases here make me think this could benefit from a pattern we've been applying using Go interfaces. The pattern encapsulates behavior (would be validation + requests in this case) such that switch cases can be removed from the code. Would be happy to pair with you to show what it looks like and see if it would work out. 

Can do this as a follow-up refactor of course, if you're time constrained.

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Yep, it would look something like [this](https://github.com/hashicorp/hcp/pull/143/files#diff-f00de939fa49d6f60296361e26cfa1dbca8436801089617b40f173ca64865310R222-R226). That's a great point, I think grouping fields (like `role_arn` in `assume_role`) would be great for organizing and even unmarshaling, but I wonder if it feels a bit unnecessary while there are so few fields atp? 

```
version: 1.0.0
type: "aws"
rotation_integration_name: "Aws-Int-12"
details: 
  default_ttl: "30"
  assume_role:
    role_arn: "ra"
```

....Then again, we would probably would want to introduce that structure now and build on that rather than wait til late when additional fields are added.

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: I think this is possible, but would take some massaging. Right now the secret config is
```
type SecretConfig struct {
	Version         string
	Type            integrations.IntegrationType
	IntegrationName string `yaml:"integration_name"`
	Details         map[string]any
}
```
with the idea of trying to keep the . There is an `UnmarshalBinary` function that can be used on proto messages, but it takes a byte array. When the file is first read, it's gathered as a []byte, but there are relevant fields (like `name`) not provided in the file, and fields that wouldn't serialize over (like secret provider `type`). Kinda tricky since we've tried to keep the structure here as generic as possible so that rotating and dynamic secrets are supported, plus all providers. But I def agree that the current hardcoded configuration is pretty brittle. Let me think about this a bit more.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -288,6 +281,70 @@ func createRun(opts *CreateOpts) error {
 				return err
 			}
 		}
+	case Dynamic:
+		sc, err := readConfigFile(opts)
```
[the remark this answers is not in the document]
mercedesbh: Sorted!

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Sorted!

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Sounds cool! Got an example I can check out?

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Thought about it a bit more. I prefer your suggestion! Made [the change](https://github.com/hashicorp/hcp/pull/143/commits/12b7a5d75239dbb8691aea393e47685e2a8cbd51)

austingebauer: Thanks!

austingebauer:
Having a look now @mercedesbh!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -283,7 +278,88 @@ func createRun(opts *CreateOpts) error {
...
+		missingFields := validateSecretConfig(sc)
+
+		if len(missingFields) > 0 {
+			return fmt.Errorf("missing required field(s) in the config file: %s", missingFields)
+		}
+
+		switch sc.Type {
+		case integrations.AWS:
+
+			missingDetails := validateDetails(sc.Details, AwsRequiredKeys)
```
austingebauer, line 301: Curious if there has been any past discussion around performing validation to the level of detail (i.e., individual keys on provider-specific APIs) we have here. I would be okay with us dropping the CLI validation for fields at this level. I could imagine it being difficult to keep the API and CLI in sync over time as fields are added / removed / made optional. The APIs typically return an error for any required fields that are missing which can be raised to users here. Lmk your thoughts!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -283,7 +278,88 @@ func createRun(opts *CreateOpts) error {
...
+
+			if len(missingDetails) > 0 {
+				return fmt.Errorf("missing required field(s) in the config file details: %s", missingDetails)
+			}
+
+			var role AwsAssumeRole
```
austingebauer, line 307: Would be good to follow the same pattern for GCP but with the nested `service_account_impersonation` fields similar to AWS `assume_role so that it's consistent.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -144,9 +135,15 @@ type MongoDBScope struct {
 	Type string `mapstructure:"name"`
 }
 
+type AwsAssumeRole struct {
```
austingebauer, line 138: Any good reason to redefine the API structures when they're available via `preview_models.Secrets20231128AssumeRoleRequest` and such? I believe we could unmarshal/decode directly into them instead of redefining them unless I'm missing something.

internal/commands/vaultsecrets/secrets/create.go
austingebauer: Yep! The [`secretTypeFormatter`](https://github.com/hashicorp/hcp/blob/main/internal/commands/vaultsecrets/secrets/displayer.go#L21-L93) is an example. Instead of having switch cases for each secret type (kv,dynamic,rotating), we encapsulate the details in an interface.

I'd say don't worry about it now. Can give it a shot in a refactor once things are merged / working 👍

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Hmm, I'm wondering if this would be a bit easier to chat on in a Zoom call. There's a follow-up ticket [here](https://hashicorp.atlassian.net/browse/VAULT-29636) that will move from yaml to hcl/json, so the unmarshaling will change a bit very soon when that switches over (picking up that ticket right after this one). As for redefining structs as opposed to using the proto messages directly, seems like the complexity comes in because the bytes read from a file do not fit the schema of the proto messages perfectly. For example, the `SecretConfig` that users are expected to provide has some top level fields that won't fit provider-specific request bodies because those details are nested in the `details` field of the config file, which I'm accessing after it's been parsed. Unless I'm missing something obvious 🤔.

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Great point!

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Sounds good, I'll add this as a follow-up item!

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: This is a great point. In the RFC, we added [a section](https://hermes.hashicorp.services/document/1MR6Fu60VISKJCBJhCpyEC1VqeizJCwJeiYbX6X88bd0) for a backend endpoint that would not only help users compose the configuration files, but also give us a source of truth to work from that isn't hard-coded "required keys" in this repo. We bumped that work's priority up to Milestone 3 of this project (instead of just being "future work"! While it's still hard-coded values though, we'll run into issues places like [here](https://github.com/hashicorp/hcp/pull/143/files#diff-7f92b17a95650abeca3ec432e8bcf6859f19709d1bdb6b3687fb5d704940a4ecR324) with panics on interface conversion when the value wasn't provided in the config file (since it would be nil and not a string atp).

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Chatted a bit on a sync, we're going to be changing to hcl/json in the [next ticket](https://hashicorp.atlassian.net/browse/VAULT-29636), so we don't want to add to much work here considering it would be undone in the next PR. Because the `SecretConfig` doesn't map to the proto messages, we'll have to see what's offered (if anything) via some hcl libraries to try and unmarshal/decode directly from the hcl file to request bodies.

austingebauer (approved):
LGTM. Thanks for chatting with me about this yesterday.

PR #147, opened 2024-08-20
[HVS] Support Pagination in run Command

VinnyHC:
Implement the ability to use `next_page_token`'s to accumulate secrets from paginated responses. 

### How I've tested this PR:
- New tests have been added

- Can be tested against product Vault Secrets resources

### Checklist:

### DIFF
--- a/internal/commands/vaultsecrets/run/run.go
+++ b/internal/commands/vaultsecrets/run/run.go
@@ -142,8 +142,6 @@ func getAllSecretsForEnv(opts *RunOpts) ([]string, error) {
 	secs, err := fetchPaginatedSecrets(opts)
 	if err != nil {
 		return nil, err
-	} else if len(secs) == 0 {
-		return nil, fmt.Errorf("no secrets returned from app")
 	}
 
 	result := os.Environ()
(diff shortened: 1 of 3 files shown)


internal/commands/vaultsecrets/run/run.go
```diff
@@ -139,20 +139,17 @@ func runRun(opts *RunOpts) (err error) {
...
-	params.OrganizationID = opts.Profile.OrganizationID
-	params.ProjectID = opts.Profile.ProjectID
-	params.AppName = opts.AppName
-
-	res, err := opts.PreviewClient.OpenAppSecrets(params, nil)
+	secs, err := fetchPaginatedSecrets(opts)
 	if err != nil {
 		return nil, err
+	} else if len(secs) == 0 {
+		return nil, fmt.Errorf("no secrets returned from app")
```
VinnyHC, line 146: I think this is a reasonable UX improvement but happy to remove if folks don't think its needed

emoncuso: Is there a story where you'd want to "run" even if the list is empty?

VinnyHC: Yeah great question. I suppose this is a breaking change and probably best removed.

I can think of a situation where you push app code that no longer needs secrets and you delete them from HVS but your tooling is still launching the app via `hcp vs run`

VinnyHC: Removed in 526b17d , we can add later if we want it.

PR #151, opened 2024-08-26
CVE Updates

jasonpilz:
* 2 CVE's from alpine linux 3.20 have caused the docker image artifacts to fail security scans.
* Example: https://github.com/hashicorp/crt-workflows-common/actions/runs/10564866097/job/29268528445

1. [CVE-2024-7264](https://security.alpinelinux.org/vuln/CVE-2024-7264) (curl 8.9.0-r0)
2. [CVE-2024-43374](https://security.alpinelinux.org/vuln/CVE-2024-43374) (vim 9.1.0414-r0)

* Both have patched versions on alpine edge (not yet in a patched version).
  * Manually pinned the edge versions in build.
* Updated the build to base the production image on the dev build stage since it already shared the same foundational build steps.

### How I've tested this PR:
```
make docker/dev

docker buildx build \
                --load \
                --platform linux/arm64 \
                --tag docker.io/hashicorp/hcp:latest-195e341 \
                --target=base \
                .

docker buildx build \
		--load \
		--platform linux/arm64 \
		--tag docker.io/hashicorp/hcp:latest-195e341 \
		--target=dev \
		.
```

### Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: BUSL-1.1
 
-## Dev DOCKERFILE ##
-FROM alpine:3.20 as dev
-
-COPY bin/hcp /bin/hcp
+## Common Base Layer
+FROM alpine:3.20 as base
 
 RUN apk --no-cache upgrade && apk --no-cache add \
 	bash \
@@ -16,12 +14,18 @@ RUN apk --repository=https://dl-cdn.alpinelinux.org/alpine/edge/main --no-cache 
   vim=9.1.0678-r0 \
   curl=8.9.1-r1
 
-RUN touch ~/.bashrc && hcp --autocomplete-install
+RUN touch ~/.bashrc
+
+## Dev DOCKERFILE ##
+FROM base as dev
+
+COPY bin/hcp /bin/hcp
+RUN hcp --autocomplete-install
 
 CMD ["/bin/bash"]
 
 ## DOCKERHUB DOCKERFILE ##
-FROM dev as release
+FROM base as release
 
 ARG BIN_NAME
 ARG NAME=hcp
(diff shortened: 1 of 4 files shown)


internal/commands/auth/print_access_token.go
```diff
@@ -28,13 +28,6 @@ func NewCmdPrintAccessToken(ctx *cmd.Context) *cmd.Command {
 				Preamble: "To print the access token:",
 				Command:  "$ hcp auth print-access-token",
 			},
-			{
-				Preamble: "To use the access token when curling an API:",
-				Command: heredoc.New(ctx.IO, heredoc.WithPreserveNewlines()).Must(`
```
meirish, line 33: Does this get executed? It would be nice to keep it in the docs if possible.

jasonpilz: Sure, we can keep it. I was just thinking that if a user attempted this example with the docker container it wouldn't have `curl` installed, but I suppose it could just be manually installed. Will do as you suggest.

Dockerfile
```diff
@@ -3,18 +3,25 @@
...
+  vim=9.1.0678-r0 \
+  curl=8.9.1-r1
+
 RUN touch ~/.bashrc && hcp --autocomplete-install
+
 CMD ["/bin/bash"]
 
 ## DOCKERHUB DOCKERFILE ##
-FROM alpine:3.20 as release
+FROM dev as release
```
dadgar, line 24: Does this work? The build pipeline will try to build the docker target release: https://github.com/hashicorp/hcp/blob/main/.github/workflows/build.yml#L171.

The dev target, tries to copy bin/hcp to bin/hcp, but that may not exist when running the CI step. Is it safer just to duplicate the logic or to move to a script that both execute?

dadgar (approved):

Dockerfile
```diff
@@ -1,20 +1,31 @@
...
 	bash \
-	curl \
 	jq \
-	nano \
-	vim
-RUN touch ~/.bashrc && hcp --autocomplete-install
+	nano
+
+## No patch for vulnerabilities in alpine, downloading fixed versions from edge
+RUN apk --repository=https://dl-cdn.alpinelinux.org/alpine/edge/main --no-cache add \
```
pierluc-codes, line 13: I like that solution!

pierluc-codes (approved):
Good job! I like the end result.

Dockerfile
jasonpilz: Your right, thanks for pointing that out! Updated.

PR #160, opened 2024-09-12
Vault 29636 create secrets using hcl

mercedesbh:
### Checklist:


mercedesbh:
An additional patch will go up to support the recent capabilities work and the expansion to AWS and GCP rotating secrets

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -117,39 +122,16 @@ type CreateOpts struct {
...
-}
-
-type MongoDBScope struct {
-	Name string `mapstructure:"type"`
-	Type string `mapstructure:"name"`
-}
-
-type AwsAssumeRole struct {
-	RoleArn string `mapstructure:"role_arn"`
+type DetailsInternal struct {
```
dhuckins, line 125: if its internal should it be exported?
```suggestion
type detailsInternal struct {
```

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -117,39 +122,16 @@ type CreateOpts struct {
...
+type detailsInternal struct {
+	Details map[string]any
 }
 
-type GcpServiceAccount struct {
-	ServiceAccountEmail string `mapstructure:"service_account_email"`
+type SecretConfig struct {
+	Type            integrations.IntegrationType `hcl:"type"`
+	IntegrationName string                       `hcl:"integration_name"`
+	Details         cty.Value                    `hcl:"details"`
```
averche, line 132: If the `Details` was a `map[string]interface{}` instead of `cty.Value`, would the HCL parser be able to decode it correctly (eliminating the need for custom recursive decoding logic in `ctyValueToMap`)?

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -423,20 +364,56 @@ func readPlainTextSecret(opts *CreateOpts) error {
...
-		return sc, fmt.Errorf("unable to unmarshal config file: %w", err)
+		return sc, di, err
+	}
+	di.Details = detailsMap
+
+	return sc, di, nil
+}
+
+func ctyValueToMap(value cty.Value) (map[string]any, error) {
+	varMapNow := make(map[string]any)
```
averche, line 387: Nit: this map's name is a bit confusing (with `Now` suffix)

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -423,20 +364,56 @@ func readPlainTextSecret(opts *CreateOpts) error {
...
+	for k, v := range value.AsValueMap() {
+		if v.Type() == cty.String {
+			if k == "rotation_policy_name" {
+				varMapNow[k] = rotationPolicies[v.AsString()]
```
averche, line 391: I think it may be better to accept full policy name here (e.g. "built-in:30-days-2-active") instead of just "30". This will allow us to support custom policy names in the future and it would be consistent with our current API.

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: HCL[ unfortunately does not support `map[string]interface{}`](https://github.com/hashicorp/hcl/issues/291) since it requires defined types. `cty.Value` (or something like `hcl-Attributes`) is the closest thing to the flexibility we need.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -288,11 +242,10 @@ func createRun(opts *CreateOpts) error {
 		}
 
 	case secretTypeDynamic:
-		sc, err := readConfigFile(opts)
+		sc, di, err := readConfigFile(opts)
```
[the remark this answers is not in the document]
mercedesbh: Great point, sure thing

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Good catch - leftover from manual testing

averche (approved):

PR #167, opened 2024-09-27
VAULT-29635 - Create rotating and dynamic secrets from interactive cli 

mercedesbh:
For ticket https://hashicorp.atlassian.net/browse/VAULT-29635, we want to provide users with an option to create rptating and dynamic secrets through interactive CLI prompting. This PR provides that experience when users omit the `--data-file` flag.

### How I've tested this PR:
Ran `make go/build` to test manually

1. Checkout this branch
2. Run `make go/build`
3. Create a HVS rotating or dynamic secret using `./bin/hcp vault-secrets secrets create rotating_secret --secret-type=rotating` or `./bin/hcp vault-secrets secrets create dynamic_secret --secret-type=dynamic`

Screenshot 2024-09-27 at 5 57 16 PM

### Checklist:


internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -127,11 +135,70 @@ type secretConfigInternal struct {
...
 
 type SecretConfig struct {
-	Type            integrations.IntegrationType `hcl:"type"`
-	IntegrationName string                       `hcl:"integration_name"`
-	Details         cty.Value                    `hcl:"details"`
+	Type    integrations.IntegrationType `hcl:"type"`
+	Details cty.Value                    `hcl:"details"`
 }
 
+var (
```
mercedesbh, line 142: Overall, this implementation feels pretty hacky, but allowing the interactive experience from hardcoded values for now is a bit tricky. There's a function [here](https://github.com/hashicorp/hcp/pull/167/files#diff-7f92b17a95650abeca3ec432e8bcf6859f19709d1bdb6b3687fb5d704940a4ecR542) that depends on the type of the empty values in the maps below. Definitely open to an alternative solution that's either less code or not as hacky.

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+		case reflect.Slice:
+			var (
+				sliceItems []map[string]any
+			)
+			nestedMap := make(map[string]any)
+			valueSlice := value.([]map[string]string)
+
+			for proceed := true; proceed; proceed = promptForAdditionalField(field, opts) {
+
+				for _, nestedField := range maps.Keys(valueSlice[0]) {
```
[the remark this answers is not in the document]
pmmukh: also generally a little curious why you're not doing the recursive approach when the type is a map here as well, if this is a slice of maps

pmmukh (approved):
this is awesome!!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+		return config, internalConfig, err
+	}
+
+	internalConfig.Details = fieldValues
+	return config, internalConfig, err
+}
+
+func populateFieldValues(fieldMap map[string]any, opts *CreateOpts) (map[string]any, error) {
+	var fieldPrompt promptui.Prompt
+	fieldsMap := make(map[string]any)
```
averche, line 590: Nit: `fieldMap` and `fieldsMap` are a bit too similar and could get confused within the same function

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+		Stdout: iostreams.NopWriteCloser(opts.IO.Err()),
+	}
+
+	_, provider, err := providerPrompt.Run()
+	if err != nil {
+		return config, internalConfig, fmt.Errorf("provider selection prompt failed: %w", err)
+	}
+	config.Type = integrations.IntegrationType(provider)
+
+	fieldValues, err := populateFieldValues(providerFields[provider], opts)
```
averche, line 579: Is the provider here guaranteed to be a valid? (otherwise `providerFields[provider]` might return nil)

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -130,11 +138,70 @@ type secretConfigInternal struct {
...
 type SecretConfig struct {
-	Type            integrations.IntegrationType `hcl:"type"`
-	IntegrationName string                       `hcl:"integration_name"`
-	Details         cty.Value                    `hcl:"details"`
+	Type    integrations.IntegrationType `hcl:"type"`
+	Details cty.Value                    `hcl:"details"`
 }
 
+var (
+	twilioRotatingSecretConfig = map[string]any{
```
averche, line 146: Maybe calling these templates would be more appropriate since they are not meant to be modified

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+	}
+	config.Type = integrations.IntegrationType(provider)
+
+	fieldValues, err := populateFieldValues(providerFields[provider], opts)
+	if err != nil {
+		return config, internalConfig, err
+	}
+
+	internalConfig.Details = fieldValues
+	return config, internalConfig, err
```
averche, line 585: Style nit:

```suggestion
	if err != nil {
		return SecretConfig{}, secretConfigInternal{}, err
	}

	internalConfig.Details = fieldValues

	return config, internalConfig, nil
```

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+}
+
+func populateFieldValues(fieldMap map[string]any, opts *CreateOpts) (map[string]any, error) {
+	var fieldPrompt promptui.Prompt
+	fieldsMap := make(map[string]any)
+
+	for _, field := range maps.Keys(fieldMap) {
+		value := fieldMap[field]
```
averche, line 593:
```suggestion
	for field, value := range fieldMap {
```

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -439,9 +527,124 @@ func validateSecretConfig(secretConfig SecretConfig) []string {
...
+		switch reflect.ValueOf(value).Kind() {
+		case reflect.String:
+			fieldPrompt = promptui.Prompt{
+				Label: field,
+				Mask:  '*',
+			}
+
+			input, err := fieldPrompt.Run()
+			if err != nil {
+				return fieldsMap, fmt.Errorf("prompt for field %s failed: %w", field, err)
```
averche, line 603:
```suggestion
				return nil, fmt.Errorf("prompt for field %q failed: %w", field, err)
```

averche (approved):
Left a few comments above but nothing blocking!

internal/commands/vaultsecrets/secrets/create.go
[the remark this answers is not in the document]
mercedesbh: > i think i'm a little confused on the logic behind maps.Keys(valueSlice[0]) here, would you mind dropping a comment about it?
I can definitely drop a comment here. In short, since the `valueSlice` (slice of maps) is hardcoded in the provider field templates, we know there is only a map at the first index of the slice. So instead of iterating through the slice, let's just grab the first index of the slice and iterative over the map keys at that index.

> also generally a little curious why you're not doing the recursive approach when the type is a map here as well, if this is a slice of maps
Good question. Only because the [prompt](https://github.com/hashicorp/hcp/pull/167/files#diff-7f92b17a95650abeca3ec432e8bcf6859f19709d1bdb6b3687fb5d704940a4ecR624) is a bit different when it's a slice, but let me see if I reconcile that for it to also be recursive.

internal/commands/vaultsecrets/secrets/create.go
mercedesbh: Yep, it's guaranteed to be valid. The provider was chosen by the user from a selection of providers in `availableDynamicSecretProviders` and `availableRotatingSecretProviders` ([maps](https://github.com/hashicorp/hcp/pull/167/files#diff-7f92b17a95650abeca3ec432e8bcf6859f19709d1bdb6b3687fb5d704940a4ecR533-R543) from provider to fields)

internal/commands/vaultsecrets/secrets/create.go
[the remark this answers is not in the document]
pmmukh: ahhh, didn't realize `valueSlice` would only have 1 elem there if its a slice of maps, thanks for clarifying!

internal/commands/vaultsecrets/secrets/create.go
```diff
@@ -130,11 +138,70 @@ type secretConfigInternal struct {
...
+		"integration_name": "",
+		"secret_details": map[string]any{
+			"mongodb_roles": []map[string]string{
+				{
+					"collection_name": "",
+					"database_name":   "",
+					"role_name":       "",
+				},
+			},
+			"mongodb_scopes": []map[string]string{
```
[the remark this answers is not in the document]
mercedesbh: Screenshot 2024-10-09 at 3 59 05 PM

PR #184, opened 2024-10-17
VAULT-31803 - Updates secret_name to secret

mercedesbh:
The `SecretName` field has now been deprecated, and rotating secrets require the usage of `Name` instead. This PR makes those updates in accordance with [that deprecation](https://github.com/hashicorp/cloud-vault-secrets/pull/2155).

### How I've tested this PR:
Ran `make go/build` and tested locally

1. Checkout this branch
2. Run `make go/build`
3. Try out any command that has now switched from `SecretName` to `Name` (example, `./bin/hcp vs secrets create twil-sec --secret-type=rotating`)

Screenshot 2024-10-17 at 12 53 04 PM

### Checklist:

### DIFF
--- a/.changelog/184.txt
+++ b/.changelog/184.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+vault-secrets: Update vault-secrets rotating secrets from `secret_name` to `name` usage
+```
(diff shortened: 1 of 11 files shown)


averche (approved):
Looks good! Just to double check, is the CLI using any of these MongoDB [response fields](https://github.com/hashicorp/cloud-vault-secrets/pull/2227/files#diff-3b9d0c50f151ac8ab6bc650d911b878f4b7be63aed51c2572efe0cf2c3d4c53aL2245-L2248) that will be going away?

mercedesbh:
> response fields

Thanks for the reminder to double-check on this! All good though, the only usage here is secrets creating those fields into `secret_details` already, and the responses aren't being parsed at all.

Issue #215, opened 2025-02-25
`hcp vault-secrets secrets create --secret-type` does not accept the correct key


syaghoubi00:
The `--secret-type` accepts `kv` instead of `static`

```sh
echo test | hcp vault-secrets secrets create --secret-type=static --data-file=- test_secret
ERROR: "static" is an unsupported secret type; "static", "rotating", "dynamic" are available types
```

```sh
echo test | hcp vault-secrets secrets create --secret-type=kv --data-file=- test_secret
Secret Name   Created At                 Latest Version   Type
test_secret   2025-02-25T01:44:22.364Z   4                kv  

✓ Successfully created secret with name "test_secret"
```

PR #216, opened 2025-02-25
fix(vaultsecrets.secrets.create): parse correct secret-type key

syaghoubi00:
### :hammer_and_wrench:  Description

Parses correct key from `hcp vault-secrets secrets create --secret-type`

Parses the literal string `static` from flag arg

### :link:  Additional Link

Closes: #215

### :building_construction:  Local Testing

```sh
podman run --rm -it fedora
dnf install -y go
git clone ${REPO}
make go/test
make go/build
echo test | ./bin/hcp vault-secrets secrets create --secret-type=static --data-file=- test_secret
```

### :+1:  Checklist

- [x] The PR has a descriptive title.
- [ ] Input validation updated
- [ ] Unit tests updated
- [ ] Documentation updated
- [ ] Major architecture changes have a corresponding RFC


syaghoubi00:
Added changelog entry. Is anything else required?

dhuckins:
hi @syaghoubi00 thanks for the contribution!

this would mean a breaking change for anyone who uses `hcp vault-secrets secrets create --secret-type kv`, to prevent that from breaking would it be possible to allow both `"kv"` and `"static"` as acceptable to create a kv secret?

syaghoubi00:
@dhuckins I went ahead and preserved the `kv` opt. This is actually the better fix, because the `secretTypeKV` is used in more places than I found at first. This will allow the interface to the `secretTypeKV` to remain consistent in code, while still parsing `static` from the arg opts.

It might be a good idea to normalize all the opts input, since they are being parsed as string literals. eg. `=Static` and `=STATIC` are both 'valid' opts, so adding something like `input.ToLowercase` might be better for doing the case evals. 

https://github.com/hashicorp/hcp/blob/e29db0534015333e626b90fa4c35d3c37f62fdf2/internal/commands/vaultsecrets/secrets/secret_types.go#L6-L10

PR #243, opened 2025-06-19
Make Waypoint action agent `status` stanzas work

jgwhite:
### :hammer_and_wrench:  Description

The agent config technically supports the following:

```hcl
group "my_group" {
  action "my_action" {

    operation {
      status {
        message = "Done"
        status = "complete"
      }
    }

  }
}
```

However, this will crash with a null pointer exception because the status operations expects a Waypoint client to be available here:

https://github.com/hashicorp/hcp/blob/ce2e5e40abfa2246690a0454c7871fc8b6ae8a4a/internal/pkg/waypoint/agent/status_operation.go#L30-L31 

But it doesn’t look like we’re ever initialising it.

Additionally, the `SendStatusLog` RPC currently requires action ID and run sequence, which aren’t available in the `AgentOperation` type that the agent retrieves. I’ve submitted https://github.com/hashicorp/cloud-waypoint-service/pull/1718 to address this and allow us to send a status log with just a run ID.

**TL;DR this patch will:**

1. Plumb through the Waypoint API client
2. Adopt the new signature for `SendStatusLog` (when available)
3. Add tests for the `status` operation

### :link:  Additional Link

- [WAYP-3495](https://hashicorp.atlassian.net/browse/WAYP-3495)

### :building_construction:  Local Testing

1. Check out and build this branch
2. Place the `agent.hcl` listing from below somewhere on your filesystem (i.e. `~/Desktop/agent.hcl`)
3. [Visit the `jamie-test` action](https://portal.cloud.hashicorp.com/services/waypoint/actions/jamie-test?project_id=66263cbe-f18d-4d6b-955f-a01ae1096da4)
4. Click “Run” and enter a message
5. Run the Waypoint agent locally
   ```sh
   hcp waypoint agent run -c ~/Desktop/agent.hcl
   ```
6. Verify you hear your computer speak the message
7. Verify that a new run appeared in the web UI
8. Verify that the run contains two status logs (see screenshot)

<details>
<summary><b>📄 agent.hcl</b></summary>

```hcl
group "demo" {
  action "say" {
    operation {
      status {
        message = "Clearing throat"
        values = { "cough": "ahem" }
        status = "running"
      }
    }

    operation {
      run {
        command = "say '${var.message}'"
      }
    }

    operation {
      status {
        message = "Taking a bow"
        status = "complete"
      }
    }
  }
}
```

</details>

<details>
<summary><b>🖼️ Screenshot</b></summary>

CleanShot 2025-07-01 at 12 07 53@2x

</details>

### :+1:  Checklist

- [x] Await #245
- [x] The PR has a descriptive title.
- [x] Unit tests updated
- [x] CHANGELOG entry added

## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [x] If applicable, I've worked with GRC to document the impact of any changes to security controls.
  Examples of changes to controls include access controls, encryption, logging, etc.
- [x] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.
  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

[WAYP-3495]: https://hashicorp.atlassian.net/browse/WAYP-3495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changelog/243.txt
+++ b/.changelog/243.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-Waypoint: Fix Waypoint agent stanza operations
+Waypoint: Fix Waypoint agent status operations
 ```


internal/pkg/waypoint/agent/status_operation.go
```diff
@@ -5,33 +5,44 @@ package agent
...
 )
 
 type StatusOperation struct {
-	wp waypoint_service.ClientService
-
-	orgID  string
-	projID string
-	cfgID  string
-	runID  string
-
```
jgwhite, line 23: I have no idea how Evan was originally planning to populate these. Possibly some sort of HCL-parsing voodoo that I’m not familiar with.

internal/pkg/waypoint/agent/execute.go
```diff
@@ -27,7 +29,13 @@ var (
 )
 
 type Operation interface {
-	Run(ctx context.Context, log hclog.Logger) (OperationStatus, error)
+	Run(
```
paladin-devops, line 32: I agree that this works (I built this branch myself locally and tested it out 😄), however I'm unsure if updating this function in the interface is the right approach here. It appears that `StatusOperation` is the only type of operation that (currently) uses `api`, `profile`, or `opInfo`. That's not to say that others won't in the future, but I feel like the dependency injection that was laid out initially by including `waypoint_service.ClientService` as a struct field on `StatusOperation` is the better approach.

That still leaves the issue of when/where to initialize the client field, because as you correctly pointed out in the ticket/PR description there is a nil pointer dereference happening as it's not initialized right now, prior to running the operation [here](https://github.com/hashicorp/hcp/blob/a5fb60bffbc173d88542a2eac9751b7dca8da609/internal/pkg/waypoint/agent/execute.go#L86). I think this should therefore be done in `convertAction` [here](https://github.com/hashicorp/hcp/blob/a5fb60bffbc173d88542a2eac9751b7dca8da609/internal/pkg/waypoint/agent/config.go#L135), which would also mean passing it through `Action` [here](https://github.com/hashicorp/hcp/blob/a5fb60bffbc173d88542a2eac9751b7dca8da609/internal/pkg/waypoint/agent/config.go#L99) first. Then, in `convertAction` for the case of a `status` operation, `convertStatus` would set up the necessary fields.

On the other hand, I don't want this to be a blocking change request, if we think that the other operation types will make use of these fields in the future; because in that case, including it on the interface would mandate their availability for such future changes. Happy to chat more about this. Seeing the unused input fields by several "concrete" types implementing the interface is what gave me pause here to go down recommending this change.

HenryEstberg: Good callout here Joe. The additional unused fields do strike me as a little bit odd, but I think it is not necessarily a dealbreaker for me as I could see us using the clientService elsewhere in the future.

I like the idea of including the `waypoint_service.ClientService` on `StatusOperation`. I'm a little bit torn on moving the initialization of the fields inside of `convertStatus` and `convertAction` because to me that feels quite complex and like we would be making the codebase harder to maintain in the future. For an engineer with less context, the approach of updating the interface itself to carry the information seems more straightforward. 

I do think that either approach is fine if clearly documented, but I am drawn to what seems like a simpler approach in changing the interface. If this is a pattern we want to avoid however I am willing to change my opinion.

paladin-devops: I think we can move forward with this as-is. I think we should open a ticket to look into refactoring this part of the agent code down the road, as a preference to adding more input parameters to the `Run` function in the future. I don't know or anticipate that we would, but I think that perhaps doing the "constructor" part of the work that the convert functions I referred to are doing should be separated from the "conversion" work being done, and thus make doing it that way simpler. But again, I would rather see this shipped in its current form to fix the bug than see that freshened up right now. 😄

